PR #21, opened 2022-06-09
Create readonly mirrors

caendesilva:


caendesilva:
Ignore the reopen, I just wanted to test if it was possible to restore very old branches

PR #38, opened 2022-06-15
Add DevTools package skeleton 

emmadesilva:


caendesilva:
Closing. It's better to handle this in an autoload-dev of Hyde/Framework

caendesilva:
Here's a link to the only command added https://github.com/hydephp/develop/blob/2af26e2007abb09b755c420e8962ec9c5c69489d/packages/devtools/src/Commands/MakeCommandCommand.php

PR #73, opened 2022-06-20
Remove deprecated CreatesDefaultDirectories action

caendesilva:
The action has been present in Hyde for a long time, originally added to create the default directories required by the Application automatically. To prevent any issues with file generations, this action was called through the service provider whenever the application is booted which is of course very inefficient. Furthermore, once Hyde started to support changing the source and output directories this whole system began to be less reliable. Since then we've moved to dynamically create directories as they are needed.

This PR also adds some .gitkeeps primarily to give a good overview of the Hyde directory structure.


caendesilva:
Okay, all the actual checks are passing. But for some reason codecov analysis is failing. I think it may be because a test was removed?

PR #106, opened 2022-06-24
Add a safer output directory config option to replace deprecated setting

caendesilva:
## About
The `hyde.site_output_path` option was recently deprecated. This update adds an alternative which resolves the relative path within the service provider. This makes it safer to set the output directory to something within the project. If one wants to use an absolute path outside the project that must be done in a service provider, which has now been documented.

> **Warning**
> Hyde deletes all files in the output directory before compiling the site. Don't set this path to a directory that contains important files!

If you want to store your compiled website in a different directory than the default `_pages`, you can change the path using the following configuration option in config/hyde.php. The path is expected to be relative to your project root.

```php
// filepath config/hyde.php
return [
    'output_directory' => 'docs',
];
```

### Setting an absolute path 
If you want to store the output website outside your project with an absolute path you may do so at your own risk using a service provider. This is not supported or recommended as it may cause unintentional file deletions.

```php
// filepath Boot method of a service provider
StaticPageBuilder::$outputPath = '/var/www/my-project/';
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This serves two purposes:
 
 ### Added
 - Added configuration option `hyde.media_extensions` to allow you to specify additional comma separated media file types. https://github.com/hydephp/develop/issues/39;
+- Adds a safer config option `hyde.output_directory` for customizing the output directory.
 
 ### Changed
 - for changes in existing functionality.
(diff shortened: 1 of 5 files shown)


caendesilva:
Think it's okay that we lost some coverage since the deprecated method will be removed soon

PR #112, opened 2022-06-25
Implement router and improve models

caendesilva:
There are two intentions behind creating a router. One, to provide a convenient helper for developers writing their own Blade views. And secondly, to provide a singular source of truth for the site routes.

The actual process is not changed all that much. There is still no way to manually register routes. Instead, see the new routing module as a way to strictly define how the Hyde autodiscovery works.

This will be especially useful for integrations like the Realtime Compiler that currently relies on a decoding algorithm that breaks if a source directory is customized. Here the integration could boot the Hyde application and get the processed route collection, or it could be pre-compiled into a JSON file similar to how source mapping works. 

In fact, I think it may be better to think of this all as source maps that link together page models with source files and output files as well as URI paths into an interconnected web, or net. Hmm, that reminds me of something.

The update also adds a few new interfaces and methods to define the structure for as well as store information about where files are found and where they should go. This also has the added benefit that site output directories will be supported by all page models. Want to store blog posts in the blog subdirectory? No problem.

TODO:

- [ ] Add tests (right now I actually have no idea if this even works as it is all just sketches at this point)
- [ ] Write documentation (tl;dr: if you are not messing with Blade and PHP integrations you don't need to worry about this, as the automatic routing is still the same, (with the exception that it may be impossible to get route naming conflicts which is a plus and we can remove that validator))
- [ ] Update changelog


caendesilva:
Think some of the changes here are premature abstraction and can be simplified a lot.

caendesilva:
Closing in favour of the merged https://github.com/hydephp/develop/pull/168

PR #114, opened 2022-06-25
Push split monorepo directly to master branches

caendesilva:
Want to test pushing the monorepo packages directly to the master branches. Since we are in v0.x I think it's okay if they are not entirely stable. This will reduce developer complexity a lot, and make it faster to deploy releases, allowing for more frequent and more scoped releases with less headache as merge pull requests don't need to be synced.

Once Hyde enters 1.x and becomes stable develop branches would make more sense.

### DIFF
--- a/.github/workflows/split-monorepo.yml
+++ b/.github/workflows/split-monorepo.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: hydephp/framework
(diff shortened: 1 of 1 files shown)


caendesilva:
Got an idea that an easy way to have a stable (as stable as it gets during beta) branch to push changes to master, but upon release the release changes are pushed to a stable branch

Issue #169, opened 2022-07-04
Move the build loop to a service


caendesilva:
I like keeping console commands light, as I think they should mainly be an easy way to access and trigger services. Think of it as the frontend to an API. Currently the main Hyde static site building process is done in the HydeBuildStaticSiteCommand, and while I managed to put a lot of the logic into traits I do still think the command class can be simplified by moving to a service, which would also use the new router.

The reason I've had everything in the command is because I did not want to lose the Symfony progress bars, but I'm rather sure that we can access those by passing the output interface to the service.

PR #174, opened 2022-07-04
Create build service to utilize the router and move logic away from command 

emmadesilva:
This update refactors the static site building process to handle logic in a service, to fix https://github.com/hydephp/develop/issues/169.

As part of this refactor, the build loop no longer pulls data from the collection service, instead, the route index provided by the new Router serves as the origin of truth for what files to compile.

The rebuild service has not yet been refactored.

### DIFF
--- a/packages/framework/src/Services/BuildService.php
+++ b/packages/framework/src/Services/BuildService.php
@@ -57,7 +57,7 @@ class BuildService
         $this->needsDirectory(Hyde::getSiteOutputPath('media'));
 
         $collection = CollectionService::getMediaAssetFiles();
-        $this->comment("Transferring Media Assets...");
+        $this->comment('Transferring Media Assets...');
 
         $this->withProgressBar($collection,
             function ($filepath) {


packages/framework/src/Services/BuildService.php
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    protected function getDiscoveredModels(): Collection
+    {
+        return $this->router->getRoutes()->map(function (Route $route) {
+            return $route->getPageType();
+        })->unique();
+    }
+
+    protected function canRunBuildAction(array|\Countable $collection, string $pageClass, ?string $verb = null): bool
```
caendesilva, line 77: Using the router instead of the collection service brings a side effect; no empty collections will be passed to the build loop. While this sadly means we won't get the `No $name found. Skipping...` message, we can also reduce complexity by removing this helper method.

caendesilva: Resolved in https://github.com/hydephp/develop/pull/174/commits/301b904888b87bc4155a9cdcd9263930be24070c

caendesilva: Remember to update the changelog!

Issue #179, opened 2022-07-04
Convert navigation menu generation to service


caendesilva:
I think I will convert the navigation menu to a service, taking it's base from the router.

_Originally posted by @caendesilva in https://github.com/hydephp/develop/issues/76#issuecomment-1174074817_

Issue #77, opened 2022-06-20
Feature request: Add front matter option to Markdown pages to hide them from the navbar


caendesilva:

caendesilva:
I think this could be something to add as a public property in the Markdown page models added upon parsing the source files. They can then be filtered when creating the navigation menu. See https://github.com/hydephp/develop/issues/179

caendesilva:
Upcoming release will add the following front matter option to exclude a page from the navigation menu

```yaml
hidden: true
```

PR #187, opened 2022-07-05
Convert navigation menu generation to service 

caendesilva:
Will resolve the following

 

- [ ] Fix https://github.com/hydephp/develop/issues/77
- [ ] Fix https://github.com/hydephp/develop/issues/179
- [ ] Fix https://github.com/hydephp/develop/issues/193
- [ ] Fix https://github.com/hydephp/develop/issues/197


caendesilva:
Dropping support for automatic linking to readmes in documentation pages. It adds a lot of complexity, and if your docs site doesn't have an index.html page that's gonna cause larger issues with the site.

packages/framework/src/Modules/Navigation/NavigationMenu.php
```diff
@@ -0,0 +1,51 @@
...
+                $this->push($route);
+            }
+        });
+
+        return $this;
+    }
+
+    protected function getHomeRoute(): Route
+    {
+        return Route::get('index') ?? throw new RouteNotFoundException('index');
```
caendesilva, line 49: I'm a bit torn on this. On the other hand, I like to know when things go wrong. But in some cases, you may be using Hyde to just generate single pages, granted you might not need a navigation menu, but you could if you wanted external links for example. It could be nice to send it as a warning, or make it configurable, but I don't think that complexity is warranted right now.

PR #201, opened 2022-07-09
Add new support for custom navigation menu links

caendesilva:


packages/framework/src/Models/NavItem.php
```diff
@@ -23,29 +25,49 @@ class NavItem
...
         $this->title = $title;
         $this->priority = $priority;
         $this->hidden = $hidden;
     }
 
-    public static function fromRoute(RouteContract $route): static
+    /**
+     * Create a new navigation menu item from a route.
+     */
+    public static function toRoute(RouteContract|string $route, ?string $title = null, ?int $priority = null, ?bool $hidden = null): static
```
caendesilva, line 49: I don't see a use case where this would be needed. All top level routes are automatically discovered, and there are now (better and other) ways to change priorities and labels. The only time it'd be used is when adding a link to a non-root, non-documentation-index page. And that edge case I don't think warrants the added complexity. Plus, that could easily be solved by passing an href, or by creating a fake page that redirects to a nested page.

caendesilva: Resolved in #54004e60f

caendesilva: Note that Hyde will still accept routes being set in the config, but you won't be able to set titles and priorities, and it's not "officially" supported in the terms it's something we at this stage will make an effort to keep working and have tests for.

PR #211, opened 2022-07-10
Update sitemap generator to use the router

caendesilva:
Refactors how the sitemap is generated, by utilizing the router. The actual XML result does not change.


caendesilva:
I notice it is much slower (relatively speaking). About 70ms instead of 5ms when running `php hyde build:sitemap.` But when using `php hyde build` it's around 7ms which is pretty much exactly the same as the old implementation. The reason for this is that Hyde needs to discover all the routes, which takes a bit. But it only needs to do to one per application lifecycle.

PR #251, opened 2022-07-19
Breaking: Refactor documentation layouts to use Tailwind and Alpine.js

caendesilva:
As there are a lot of changes, here is a quick overview of the major ones. 

- Alpine.js is now used for interactions.
- HydeFront has been rewritten and is now on version 2.x.
- The hyde.css and hyde.js files have now for all intents and purposes been merge into app.css and refactored to Alpine.js, respectively.
- The documentation pages are now styled using TailwindCSS instead of Lagrafo.

Note that the goal with this release is to make the framework more stable and developer friendly, but without it affecting the end user experience. For example, the visual experience as well as the interactions of the refactored documentation pages are minimal and most users won't notice any change. However, for developers, the changes are significant and will reduce a lot of complexity in the future.


caendesilva:
Should not be merged until we have Cypress tests in place to cover multiple test cases to make sure it doesn't break anything.

caendesilva:
Changing scope of HydeFront, some things like specific documentation page layouts don't make much sense with TailwindCSS. Instead, I'm going to try use HydeFront as modules importable with Laravel Mix.

PR #254, opened 2022-07-19
Add Laravel Dusk for browser testing

caendesilva:
Cypress is nice and gives instant visual feedback, however it breaks productivity having to swap to a completely different framework and language. This PR will explore using Laravel Dusk, using a ported test case so it can run with Laravel Zero.

### DIFF
--- a/packages/framework/src/HydeServiceProvider.php
+++ b/packages/framework/src/HydeServiceProvider.php
@@ -120,7 +120,7 @@ class HydeServiceProvider extends ServiceProvider
         $this->app->register(Modules\DataCollections\DataCollectionServiceProvider::class);
 
         if (env('DUSK_ENABLED', false)) {
-            $this->app->register(\Laravel\Dusk\DuskServiceProvider::class);
+            $this->app->register(\Hyde\Testing\Browser\DuskServiceProvider::class);
         }
     }
 }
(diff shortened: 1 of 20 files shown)


packages/framework/src/HydeServiceProvider.php
```diff
@@ -120,7 +120,7 @@ protected function registerModuleServiceProviders(): void
         $this->app->register(Modules\DataCollections\DataCollectionServiceProvider::class);
 
         if (env('DUSK_ENABLED', false)) {
```
caendesilva, line 122: This feels very hacky to me. Code for testing should not be in a production class. If using Dusk proves useful, it's better to create a package that bootstraps test providers if that is needed.

packages/framework/src/HydeServiceProvider.php
```diff
@@ -120,7 +120,7 @@ protected function registerModuleServiceProviders(): void
...
-            $this->app->register(\Laravel\Dusk\DuskServiceProvider::class);
+            $this->app->register(\Hyde\Testing\Browser\DuskServiceProvider::class);
```
caendesilva, line 123: The reason a custom service provider is needed is because the default Dusk provider expects the default Laravel routing which we don't have.

packages/framework/src/HydeServiceProvider.php
caendesilva: This is also why codecov patch is failing.

PR #263, opened 2022-07-23
Refactor the Hyde facade into service container singleton

caendesilva:
Currently, the entire Hyde facade (`Hyde\Framework\Hyde`) is based on static methods and properties. This update will refactor to put Hyde into the Laravel service container.


caendesilva (commented):
This will also make it nice and easy to swap out the kernel for a customized one.

caendesilva (commented):
A test to ensure the singleton state is persisted across various access methods would be good.

PR #270, opened 2022-07-24
Add Alpine.js

caendesilva:
[Alpine](https://alpinejs.dev/) is a rugged, minimal tool for composing behavior directly in the markup.

Adding it to HydePHP will allow us to port HydeFront scripts directly to the Blade views, allowing for extensive customization, and improved compatibility between versions, as well as reduced complexity.


caendesilva (commented):
Added Alpine.js to upcoming release notes

PR #285, opened 2022-07-28
Add support for nested source files

caendesilva:


caendesilva:
First, the documentation sidebar needs to be refactored to use the routing system. And possibly also extend/utilize the navigation menu service.

PR #286, opened 2022-07-29
Refactor documentation sidebar to use routes

caendesilva:
This PR refactors how the documentation sidebar is generated. The main class responsible for generating it is now extending the base navigation menu, and the sidebar items are now NavItems. This makes it more maintainable as overlapping code is reduced. It also means we properly utilize routes for the sidebar.


packages/framework/src/Models/NavigationMenu.php
```diff
@@ -23,7 +23,7 @@ public function __construct()
 
     public static function create(?RouteContract $currentRoute = null): static
     {
-        return (new self())->setCurrentRoute($currentRoute ?? Hyde::currentRoute())->generate()->filter()->sort();
+        return (new static())->setCurrentRoute($currentRoute ?? Hyde::currentRoute())->generate()->filter()->sort();
```
caendesilva, line 26: Fix unsafe usage of new static

PR #287, opened 2022-07-29
Rename sidebar views

caendesilva:
Some of these component names were really long, and overly specific. We don't need to point out that the sidebar component is a "sidebar navigation menu" -- it's not like we have other types of sidebars.


caendesilva:
I do think I want to refactor the entire docs layout to be more component based, but that's for another release.

PR #292, opened 2022-07-30
Add new URL path helpers

caendesilva:
Having a function that can return both a string and a psuedo-boolean (false) can lead to unpredictable states. This PR deprecates the `Hyde::uriPath()` helper and adds two new ones instead, `Hyde::qualifiedUrl()` and `Hyde::hasSiteUrl()`, effectively splitting the former into the latter.


caendesilva:
Since the refactor to fully use the new helper will involve the article views I first want to add tests for those.

PR #297, opened 2022-07-30
Refactor Kernel to extract related code into Foundation classes

caendesilva:
> "When one class does the work of two, awkwardness results."

The goal of this experiment is to extract related code from the HydeKernel, without changing end usage. I'm doing this by creating new foundation classes, and redirecting method calls to them.

This is a rather weird pattern, but it separates logic by concerns, without changing the user experience, which I like.

If this is something that works out, each foundation class should have a test, and a contract (that both the class and the kernel implement)


caendesilva:
Since all the tests are so scattered into unit tests it's hard to migrate all of them, so I expect there to be some mess as I can't really migrate everything right now.

PR #308, opened 2022-07-31
Handle page compiling within the page models

caendesilva:
The page models are used to hold data for how pages are compiled. So why not define the compile rules there as well? This will also make it much easier to add a dynamic framework for registering new page types, without having to hard-code them in the static page builder.

Internally, this refactor is done by adding a new `CompilableContract` interface that the page models now implement. For Markdown pages, a new static property has been added to hold the view name (making it easy to swap out).

The results for the end user does not change. However, it will make the framework more modular, and reduce complexity.


caendesilva:
Tagging this for the release after the next one, so holding off with updating the release notes.

Issue #299, opened 2022-07-30
The Hyde facade does not seem to return same object instance as Hyde::getInstance()


caendesilva:
Discovered in unrelated test, noting down so I can look into it tomorrow.

How I found it:
  
    $this->filesystem = new Filesystem(Hyde::getInstance());
    Hyde::setBasePath('/foo');
    // Hyde::path() is not '/foo'
    
    // But this works:
    Hyde::getInstance()->setBasePath('/foo');

caendesilva:
Debugging this, and these two methods don't return the same object reference

        $this->assertSame(HydeKernel::getInstance(), Hyde::getInstance());
        $this->assertSame(HydeKernel::getInstance(), Hyde::getFacadeRoot());

caendesilva:
Yeah, the first two of the following reference the same object, the third one is a separate instance

  app(HydeKernelContract::class)
  Hyde::getInstance()
  Hyde::getFacadeRoot()

caendesilva:
I suspected this had something to do with caching, but turning that off did nothing. However, overriding the getFacadeRoot method, fixed the problem. To be honest, I don't know if there's any drawbacks to this.

caendesilva:
Not sure if there is a good solution to this, but I'm also not sure how much of a problem it is. Not at this point in time at least. The only current class property we have is the path property, and you usually wouldn't change that all the time during the app lifecycle.

PR #309, opened 2022-07-31
Override facade method to return the HydeKernel instance directly from the service container 

caendesilva:
Fixes https://github.com/hydephp/develop/issues/299


caendesilva:
This method did not work.

PR #311, opened 2022-07-31
Refactor filesystem helpers

caendesilva:
Breaking changes:

- Breaking: copy helper has been simplified and no validates files
- Breaking: copy helper now always uses the project path


caendesilva:
`@method` tags needs to be updated

PR #318, opened 2022-08-01
Refactor Markdown module

caendesilva:
Moves Markdown services into the Modules\Markdown namespace, and merges files to reduce complexity.


caendesilva:
## Benchmarks

As part of this refactor I want to convert the MarkdownConverter into a singleton, as we create a new converter each time we convert Markdown, which is of course very inefficient. So I'm doing some benchmarks just for fun, to see the difference.

### Baseline

#### Parsing `Hello World!`

```
Benchmark information:
Total iterations:       1500
Total execution time:   648.8ms
Avg.  iteration time:   0.43253263ms
Avg.  iterations/sec:   2311.96
Approx. Memory usage:   18MB
```

#### Parsing 2KB [Markdown fixture](https://github.com/hydephp/develop/blob/3d89f93bf025910f5e5eb31d409697d02c913b14/tests/fixtures/markdown.md)

```
Total iterations:       500
Total execution time:   1356.74ms
Avg.  iteration time:   2.71348619ms
Avg.  iterations/sec:   368.53
Approx. Memory usage:   20MB
```

#### System information:

```
PHP version: 8.0.13 (cli)
OS/Arch:     WINNT (64-bit)
xdebug:      disabled ❌
opcache:     disabled ❌
```

caendesilva:
I want to reduce complexity, and the current Markdown service is rather complex and hard to visualize as there are so many conditionals. Instead, I think it'd be better to create separate converters for those pages that need it (like Documentation pages).

PR #321, opened 2022-08-02
Breaking refactor of Markdown based models

caendesilva:
#### Heavily refactor constructors of Markdown-based page models

Adds a new interface to the Markdown page model constructors, that expects instantiated FrontMatter and MarkdownDocument objects. Normally you would use the SourceFileParser to create the object.

This means that the constructor for all Markdown-based pages is completely changed. To use a format matching the old behaviour, you can use the `MarkdownPageModel::make` method.

#### Title property has been removed from page model constructors

The following syntax has been removed: `new MarkdownPage(title: 'Foo Bar')`
Instead, you can add it with front matter: `MarkdownPage::make(matter: ['title' => 'Foo Bar'])`

#### Markdown pages now have front matter in an object instead of array

This means that instead of the following `$post->matter['title']`, you would use `$post->matter('title')`, which allows you to add a fallback like so: `$post->matter('title', 'Untitled')`


caendesilva:
Alright, tests are in the green, now to refactor to hit codecov diff target

PR #344, opened 2022-08-04
Remove title property from Markdown pages

caendesilva:
No change in behaviour as it is accessed through front matter magic method


caendesilva:
Can't be merged until Blade pages support front matter

PR #346, opened 2022-08-04
Add front matter support to Blade pages

caendesilva:
## About

Part one of this will be adding all the common methods currently in the Markdown pages. Then, write a parser to statically parse Blade files for properties. 

### Syntax

I have the following syntax suggestion, as it's actually already used in templates (and I don't want to add actual YAML blocks to Blade files)

```blade
@php($title = 'Latest Posts')
```

Which is equivalent to the following Markdown front matter
```markdown
---
title: Latest Posts
---
```

### Parsing

I think this could easily be parsed similar to how we do shortcodes, by scanning lines for patterns.

The algorithm could be roughly as follows (using `@php($foo = 'bar')` as the example)

1. Get blade file contents as text
2. Split each line into array
3. If line string starts with substring `@php($`
4. Call helper method with that line as a parameter
5. Extract the variable name and split out the value.
   -  We may need to tokenize, looking for something like `@php($`+`foo`+` = '`+`bar`+`')`, making sure to trim out whitespaces and the trailing newline
6. Add the results to the front matter

### DIFF
--- a/docs/architecture-concepts.md
+++ b/docs/architecture-concepts.md
@@ -75,10 +75,17 @@ Take a look at the [customization and configuration guide](customization.html) t
 
 ## Front Matter
 
+### About
+
+Front matter is heavily used in HydePHP to store metadata about about pages. Hyde uses the front matter data to generate rich and dynamic content. For example, a blog post category, author website, or featured image.
+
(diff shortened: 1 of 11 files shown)


caendesilva:
Got it implemented pretty nicely, now trying to figure out how to parse arrays so they can be converted to YAML.

caendesilva:
Docs need an update

Issue #258, opened 2022-07-22
Write a YAML schema for all the currently supported front matter properties for blog posts


caendesilva:

caendesilva:
Had a random thought, what if we could write schemas as traits? That way the properties could be strongly typed and directly enforced. And the front matter when accessed could merge those properties into the matter array at runtime.

caendesilva:
Forgot the schema is actually documented, so this is gonna be a lot easier https://hydephp.com/docs/master/blog-posts#supported-front-matter-properties

caendesilva:
Maybe also add a note that the slug/identifier is reserved by Hyde and cannot be overridden by front matter

caendesilva:
Thinking a bit on how I want to handle these. On one hand, I like the idea of having these traits to only house data like follows (with comments removed for brevity)
```php
trait BlogPostSchema
{
    public string $title;
    public ?string $description = null;
    public ?string $category = null;
    public ?DateString $date = null;
    public ?Author $author = null;
    public ?Image $image = null;
}

```

But being able to construct and configure the data right there would be nice, at least when developing, so I'll try that.

caendesilva:
Also want schema for general pages to include the common shared data and any meta tags and opengraph tags. The whole point of this change is to make it clear to users and developers what data is accessible and where it is generated from.

caendesilva:
Upside is also that it should be easy enough to write a parser that automatically makes changes to the docs based on the actual schema being used (not that that will be an issue later on in v1 as a schema change is probably breaking)

PR #351, opened 2022-08-05
Change internal page data structure to use new front matter schema traits

caendesilva:
This PR adds schemas as described in #258. These schemas define the public facing data shared to the views, intended to be used by the page templates. For example metadata and SEO information.

The schemas are defined using PHP traits, that also hold the actual data, and may also construct them.

```php
trait BlogPostSchema
{
    public string $title;
    public ?string $description = null;
    public ?string $category = null;
    public ?DateString $date = null;
    public ?Author $author = null;
    public ?Image $image = null;
}

class MarkdownPost
{
    use FrontMatter\BlogPostSchema;
}
```

Another benefit of this is that we are also less reliant on magic methods and properties. Those offered through the front matter object would only really be needed for ad-hoc data added by the user.

There are more schemas to add than what will be added in this PR. Such as documentation pages and image objects. Everything that can be supplied as front matter should be covered.


caendesilva:
The image constructors are a bit of a mess so I didn't migrate all the tests as I want to do some pretty heavy refactoring

PR #355, opened 2022-08-06
Add more front matter schema traits and refactor page data structures

caendesilva:
Adds missing schema for documentation pages, and also adds some documentation in a draft

It also removes the pagemodelconstructor class as is now no longer needed
It also removes the AbstractPage magic methods as they are causing more trouble than they're worth, and we don't need them as much now with the schemas.

> This means that if you want to access something that truly is in the front matter, call the getter on the matter property. My intention is for that however, to be used as a data origin for constructing dynamic data which is then used by these prioritizing page matter methods.


caendesilva:
Now that the complex constructor has been removed we can revert the static make constructor or merge it into the constructor directly

caendesilva:
We need tests to make sure front matter set yaml values take precedence over dynamic data

PR #367, opened 2022-08-07
Refactor metadata system

caendesilva:
Refactors the metadata system. A major change is that page metadata is now stored as a page property, making it easier to see, understand, and debug. The biggest change however, is that the metadata is now strongly typed with new metadata models. Furthermore, all metadata generation is centralized so it's easier to manage.


caendesilva:
We must make sure this does not break relative links. We may need to force a recompute of relative links each time the current page property is set.

caendesilva:
Did a quick git diff test by comparing compiled site from master and this branch using the same fixtures.

The only differences I see is that now the tags are sorted by group, which should not change semantics. The comment about blog post specific tags is also removed.

packages/framework/src/Models/Metadata/Metadata.php
```diff
@@ -0,0 +1,143 @@
...
+        return $array;
+    }
+
+    protected function resolveImageLink(string $image): string
+    {
+        // Since this is run before the page is rendered, we don't have the currentPage property
+        // but since we know that this is for a blog post we know what the property will be
+        // since Hyde does not currently support custom Blog post output directories.
+        return str_starts_with($image, 'http') ? $image : "../$image";
+    }
```
caendesilva, line 142: A better way to resolve this issue that's caused by pre-computing relative data is to have the property link to a new abstract route that is not related to a page model. That way it can be resolved at runtime.

caendesilva:
Ran some benchmarks, and while the page parsers take longer, the static site builder has a marginal difference. This is expected, since we moved when we calculate data (to the parsers instead of builder). So it all evens out.

PR #381, opened 2022-08-08
Add a default category property value for documentation pages

caendesilva:
This is more predictable than assigning "other" at runtime. The logic determining if the categorized layout should be used has been updated, so this PR should have no effect on the actual rendered pages and their behaviour.


caendesilva:
Was thinking if we should make the default category name configurable (as the implicit option through the Blade view is now removed). But honestly not sure if it's worth the complexity since I think most sites using categories probably won't leave some empty. And I can't think of another title that would really suit better.

PR #395, opened 2022-08-09
Refactor internal routing structure to decouple page index from route index

caendesilva:
This PR restructures how discovered page objects are stored. The route index and page collections will now be stored as properties of the HydeKernel. The route models will be decoupled and no longer house the page objects, but instead, references to them. A collection class is added to house the actual parsed pages.

The RoutingService has been removed, and is replaced with RouteCollection.


packages/framework/src/Contracts/AbstractPage.php
```diff
@@ -80,6 +84,8 @@ public static function files(): array|false
     /** @inheritDoc */
     public static function all(): Collection
     {
+        // @todo Get from the PageCollection class.
```
caendesilva, line 87: Resolve todo before merge

packages/framework/src/PageCollection.php
```diff
@@ -0,0 +1,79 @@
...
+        if (Features::hasDocumentationPages()) {
+            $this->discoverPagesFor(DocumentationPage::class);
+        }
+
+        return $this;
+    }
+
+    protected function discoverPagesFor(string $pageClass): void
+    {
+        // @todo Parse the pages here
```
caendesilva, line 64: Resolve todo before merge

PR #451, opened 2022-08-31
Bugfix for search index generator: estimation message should only be shown when exceeding 1 second

caendesilva:


caendesilva:
Think I want to refactor this whole process

PR #453, opened 2022-08-31
Refactor auxiliary build services and tasks

caendesilva:
This update refactors the build services for the generation of sitemaps, RSS feeds, and documentation search index. These are now handled by post-build tasks, and are simplified.

Much of the return information has been removed as well as preflight checks. We may want to add some exceptions in the related services as the following messages (among others) have been removed

- Cannot generate sitemap.xml, please check your configuration.
- You don't have a site URL configured. Check config/hyde.php
- You have disabled sitemap generation in config/hyde.php
- You can enable sitemap generation by setting `site.generate_sitemap` to `true`
- Cannot generate sitemap.xml, please check your configuration.
- You don't have the `simplexml` extension installed. Check your PHP installation.

I also had to remove the output message tests as they stopped working, but they are present in the terminal when running builds. See https://github.com/hydephp/develop/commit/19d3e7db5cd9d709fbb34861f6761afcdd3022d7

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -16,7 +16,7 @@ This release contains breaking changes regarding the PostBuildTasks that may req
 - AbstractBuildTask::handle and BuildTaskContract::handle now returns null by default instead of void. It can also return an exit code.
 
 ### Deprecated
-- for soon-to-be removed features.
+- Deprecated ActionCommand.php as it is no longer used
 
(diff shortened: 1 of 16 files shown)


caendesilva:
Coverage is in large part failing due to there not being any coverage for the deprecated test, that's fine for this PR as it will be resolved in the upcoming release that removes it.

Issue #452, opened 2022-08-31
Decouple full page search page to be a Blade page stored in subdirectory


caendesilva:
This will make it more predictable and intuitive where the search.html page is coming from

caendesilva:
And a publishing option too. We can also add a validation check with a message that it can be published. Though I want to be cautious to not spend too much time on such a relatively minor feature at this point in time.

caendesilva:
The documentation will need an update https://hydephp.com/docs/master/documentation-pages#search-feature

caendesilva:
Decided against this for now. Too much complexity for to little a feature.

PR #456, opened 2022-08-31
Decouple full page search page to be a Blade page stored in subdirectory

caendesilva:
This will make it more predictable and intuitive where the search.html page is coming from. See https://github.com/hydephp/develop/issues/452


caendesilva:
Decided against this for now. Too much complexity for too little a feature.

PR #529, opened 2022-09-14
Add inline styling to front matter for post images

caendesilva:
---
    image: 
      path: cover.png
      style: 'max-height: 100px'
    ---


caendesilva:
I originally added this as a feature request because I wanted to set a max height to crop an image. I should just have cropped the image. There are very few reasons to use one-off inline styles. Modify your images beforehand, or publish the image.blade.php component to add custom classes. I considered adding a class property, but again, the point of the component is to have a consistent presentation, so why use a one-off style?

PR #548, opened 2022-10-01
Rewrite the build manifest schema

caendesilva:
## About
Adds the date the manifest was compiled, and removes the page property, replacing it with the source and output paths. Previously, the page property contained the source path. To differentiate the date from the pages, the pages are now in under the `pages` array key.

### Previous JSON format
```json
[
  {
    "page": "_pages/404.blade.php",
    "source_hash": "9a4e236420cd01cbb726240199d0de82",
    "output_hash": "c4ef14f1fa794d3d105b6d039b28ba06"
  },
  {
    "page": "_pages/index.blade.php",
    "source_hash": "72469ce3453f5b396402e5207493921b",
    "output_hash": "72469ce3453f5b396402e5207493921b"
  }
]
```

### New JSON format

```json
{
  "generated": "2022-10-01T13:40:09.602307Z",
  "pages": [
    {
      "source_path": "_pages/404.blade.php",
      "output_path": "404.html",
      "source_hash": "27a3012eacef5d0672e59ddcd9866985",
      "output_hash": "e8f804948dc9890a5659b53910f783ca"
    },
    {
      "source_path": "_pages/index.blade.php",
      "output_path": "index.html",
      "source_hash": "2611703286fbf5d7a39bb459736f9aa6",
      "output_hash": "2611703286fbf5d7a39bb459736f9aa6"
    }
  ]
}
```

_Output hash for the index page is the same as the default page is plain HTML. The output hash may be null in case the file does not exist, though since the manifest is generated immediately after the build, this shouldn't happen in production._

### DIFF
--- a/packages/framework/src/Actions/PostBuildTasks/GenerateBuildManifest.php
+++ b/packages/framework/src/Actions/PostBuildTasks/GenerateBuildManifest.php
@@ -41,6 +41,7 @@ class GenerateBuildManifest extends AbstractBuildTask
     protected function hashOutputPath(HydePage $page): ?string
     {
         $path = Hyde::sitePath($page->getOutputPath());
+
         return file_exists($path) ? md5_file($path) : null;
     }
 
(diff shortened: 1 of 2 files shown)


caendesilva:
Removed breaking from title, as the addition of the build manifest feature is still unreleased.

PR #551, opened 2022-10-02
Add Scroll Bar CSS

minhaj-313:
If it is right then please merge my PR if it is wrong then suggest me. 
if it is right please label it as  hacktoberfest-accepted

### DIFF
--- a/packages/hydefront/hyde.scss
+++ b/packages/hydefront/hyde.scss
@@ -11,3 +11,24 @@
 @use 'sass/markdown/blockquotes';
 @use 'sass/markdown/codeblocks';
 @use 'sass/utilities/alpine';
+
+
+/* Scroll bar css */
+
+::-webkit-scrollbar {
+    width: 10px;
+  }
+
+  ::-webkit-scrollbar-track {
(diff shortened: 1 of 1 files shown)


caendesilva:
Thank you for your contribution to HydePHP! Unfortunately, I'm going to have to close this pull request for a few reasons:

1. The goal of the default stylesheets are to be a minimalistic and unopinionated starting point. While I appreciate you taking the effort in using the HydePHP theme colour, that's something that's not going to be applicable for a large majority of users, as it won't fit their brands. We also want the stylesheet to be as small as possible.
2. Messing with browser defaults like scrollbars can have unintended implications, and can often lead to worse accessibility. As per the [Code of Conduct](https://github.com/hydephp/.github/blob/master/CODE_OF_CONDUCT.md), we must strive to make the HydePHP packages accessible to those with disabilities.

Don't be discouraged though, as I renamed your PR title, this will allow future users to come across this PR and they can then see your code and implement it in their own projects. HydePHP users can always add their own styles by modifying the app.css in their project.

Issue #556, opened 2022-10-08
Add helper to get the File path extension


caendesilva:
This is a great task for Hacktoberfest beginners! (Some knowledge using PHPUnit may be required)

In the `File` class we have a few helpers that allows developers to write more fluent code, for example `$file->getPath()` to get the path. I propose we add a new helper, `$path->getExtension()` which uses `pathinfo($file->path, PATHINFO_EXTENSION)` to return the extension.

Here are some ideas for how I would go about implementing this, in case you need some ideas.

1. Add a public function to [`File.php`](https://github.com/hydephp/develop/blob/master/packages/framework/src/Models/Support/File.php) named `getExtension()` that uses the [`pathinfo()`](https://www.php.net/manual/en/function.pathinfo.php) function to return the extension
2. Add tests in the [`FileTest.php`](https://github.com/hydephp/develop/blob/master/packages/framework/tests/Feature/FileTest.php) file, and verify the tests pass
3. Submit a pull request with your changes to this ([`hydephp/develop`](https://github.com/hydephp/develop/pulls)) repository 

If you're interested in giving this a shot, leave a comment so others know you are working on it. And let me know if you need any help! Good luck!

chikaldirick:
Hey! I've just posted a PR that brings `File::getExtension(): string` method: https://github.com/hydephp/develop/pull/559
I've attached a simple test that check its functionality. Please, let me know if you need some adjustments there.

caendesilva:
> Hey! I've just posted a PR that brings `File::getExtension(): string` method: #559 I've attached a simple test that check its functionality. Please, let me know if you need some adjustments there.

Merged into master! Thank you for your contribution!

PR #559, opened 2022-10-08
Add 'getExtension' method to the File class

chikaldirick:
This PR adds `getExtension(): string` method to the [File.php](https://github.com/hydephp/develop/blob/master/packages/framework/src/Models/Support/File.php) class that by using `pathinfo()` method under the hood returns extension of a file as a result.
To confirm that everything works as expected I've attached a simple test in `FileTest.php` file that check few potential use cases.
This should close issue #556


caendesilva (approved):
Looks fantastic, thank you for your contribution!

Issue #562, opened 2022-10-24
The File.php model's getMimeType() helper function should support XML in the lookup table


emmadesilva:
Here's another easy Hacktoberfest bug to fix!

Since this is targeted at beginners and I want you to learn something, I won't tell you exactly what value it should return, however I am happy to offer help if you need it.

Your mission, should you choose to accept it, is to find out the proper mime type for XML files, then submit a pull request to this repository, adding the mime type mapping to the `$lookup` array in the `getMimeType()` method of the `File.php` class.

Good luck!

LunarMarathon:
Hey, I would like to try this! 
I'm not sure if you'll be assigning first but since you've mentioned to submit a PR, I'll make the changes and submit one

caendesilva:
> Hey, I would like to try this! I'm not sure if you'll be assigning first but since you've mentioned to submit a PR, I'll make the changes and submit one

Hey! Go right ahead! I don't usually assign people, but it's great that you left a comment, so other people knows that you are working on it.

Good luck and let me know if you need any help! But most importantly: have fun!

caendesilva:
Fixed by https://github.com/hydephp/develop/pull/563

PR #563, opened 2022-10-24
Added XML support to File.php

LunarMarathon:
Solves issue #562 

Let me know if any changes needs to be made. I used appilication/xml instead of text/xml as the former was recommended. I could learn something new, thanks for the opportunity!


caendesilva (approved):
Looks perfect! You found the right mime-type; `application/xml` is indeed what I was looking for, and I'm very happy you learned something! Thank you for your contribution and have a happy Hacktoberfest!

LunarMarathon:
Thank you! Wish you the same!

PR #566, opened 2022-10-24
Bump dependencies 

caendesilva:


caendesilva:
A bit of a catch 22 here. #566 will fail without #565, and #565 will fail without #566.

PR #584, opened 2022-10-25
Simplify internal metadata API

### LINKED ISSUE #536: Remove duplicate metadata from page model
https://github.com/hydephp/develop/blob/0ef6c94fc38a3f544438eb84c9e97cdc890b67c4/packages/framework/src/Models/Support/Site.php#L43

caendesilva:
Right now we have two overlapping services handling metadata (actually three since the Site class does some logic), this is of course not good for maintainability.

This PR reduces complexity, refactors the code, and adds a new class called GlobalMetadataBag to shift those responsibilities the area of concerns for the other metadata classes. It also moves some logic from the Meta facade/helper to the Metadatabag class.

The new GlobalMetadataBag class, when filtering duplicates, will see if there is a current page being rendered, and then diff out those to fix #536.


caendesilva:
The global metadata bag inherits some helpers that are not really required, maybe we should split out those to a page metadata bag?

Also, update release notes.

PR #586, opened 2022-10-26
Refactor: Add strict types declaration to all Framework PHP source files

caendesilva:


caendesilva:
This requires a lot of refactors, which is good as it fixes undiscovered type bugs.

Wondering which looks prettiest

```php
return $this->__toString();

return (string) $this;
```

PR #589, opened 2022-10-26
Simplify NavigationData by removing ArrayObject support

caendesilva:


caendesilva:
I just remembered, we need ArrayObject so it can be accessed using front matter like this: `->get('navigation.priority')`

PR #591, opened 2022-10-26
Change most Route helpers to return null instead of throwing exceptions

caendesilva:
Refactors the Route helper, so that only the `Route::get()` helper throws an exception if the route is not found, the others now return null.


caendesilva:
Replace unnecessary type annotations with example tags

caendesilva:
Add a Route::getOrFail method instead

caendesilva:
I think this will introduce more complexity than solve problems. Closing.

PR #593, opened 2022-10-26
Breaking: Change namespace Models\Support to Support

caendesilva:


caendesilva:
Some of these are actually models, so it doesn't make semantic sense to change this. Some classes could  be moved to the helpers namespace though.

PR #600, opened 2022-10-26
Create a strong route key type

caendesilva:
This was mainly an experiment, I think I will close it right away as it adds to much complexity. The problem I want to solve is to have a canonical definition of what some types are, but they may be better suited to simply documented interfaces used as unions without implementation.

Update: Changed to simply union in the type, without using it. I think I will also simplify the class a bit since we are not directly using it.


caendesilva:
Closing in favour of adding better documentation.

PR #604, opened 2022-10-27
Refactor constructor data assignments

caendesilva:
I want it to be more immediately clear where dynamic data is assigned, so experimenting with a refactor to make this more explicit.


caendesilva:
Could be interesting to making the properties readonly, at least while developing, to make sure we only manipulate data in one place as that makes it much easier to understand how Hyde works.

PR #606, opened 2022-10-27
Extract facades and simplify public API usage

caendesilva:
Extracts facades to simplify the public code API usage, without developers having to worry as much about the underlying implementations of the entry points. Also goes over the actual implementations to simplify and normalize them.


caendesilva:
Want to make sure I'm solving the right problem with the routing. I don't think we need all these helpers.

caendesilva:
I'm actually starting to wonder if we need the Route model, it doesn't really add any new data not in the pages. We could just use the page collection instead of the route collection.

caendesilva:
While I think the namespace Hyde\Routing\Route is beautiful I don't think it's needed for just a few classes, and I don't think Hyde will get more Routing classes in the foreseeable future.

caendesilva:
With these changes, one example of changed behaviour is that the navigation brand will no longer throw an exception when attempting to link to the home route and the index page does not exist. After some consideration I think this is better, as if you don't have an index page, this would be a weird place for an exception, and also, you're gonna have bigger problems than a link not leading anywhere. Also, we have validators for just this reason.

PR #610, opened 2022-10-28
Add more factories

caendesilva:
Adds more factories. My hope with these are to create isolated classes that are easy to understand and test. To do this, I'm attempting to give them as little information as possible, in order to reduce complexity as we before this PR pass along the half-constructed page object which leads to unpredictability due to the nature of a half-finished object state.


caendesilva:
At the moment these factories are rather verbose, but I think that's okay, once they're all done it's gonna be much easier to refactor out shared code.

PR #611, opened 2022-10-28
Refactor the author model

caendesilva:
Simplifies the public API by creating an author facade and simplifying the constructors.


caendesilva:
A nice addition for the future could be to register authors within the kernel, allowing them to be declared in service providers, etc, but I don't think that complexity is warranted yet.

PR #617, opened 2022-10-28
Normalize internal image path storage

caendesilva:
Leads to a more explicit, consistant, and predictable behaviour. It also makes usage more forgiving as front matter can use both _media/ or just the filename as it will be normalized.


caendesilva:
This makes the behaviour a bit different than other page types which might actually cause more confusion than the current behaviour, maybe this is suited better for a "realPath" or "sourcePath" property, but still using the normalization as to be more forgiving.

PR #631, opened 2022-10-30
Refactor the page data factory pipeline

caendesilva:
Since the factories, by design, work on a limited known set of verbose data, we have to pass a lot of parameters around. This PR experiments with adding a type of DTO called the CoreDataObject which is not a great name, but once I or someone else comes up with a better one it's easy to refactor.

Anyhow, this immutable DTO contains only data known at the earliest point in construction solving the problem of factories needing to depend on data that might not be generated yet which is the reason for having the scoped private data as it is now.

In summary, this PR doesn't change the fundamental design of the page factory data pipeline, it just aims to reduce boilerplate by having a single well defined data object that can be passed around to subfactories more easily.


caendesilva:
FeaturedImageFactory doesn't need to use the CDO as it only needs to know about the front matter.

PR #651, opened 2022-11-07
Refactor Markdown classes

caendesilva:
Refactors Markdown related classes as a lot of these are pretty much as old as the framework itself.

Notable changes:
- [Breaking: Rename shared data property $markdown to $content
](https://github.com/hydephp/develop/pull/651/commits/1ae5d1be1ec6e2c7f7cb095f2683b8fd6d88bc40)
- The SemanticDocumentationArticle class has been heavily refactored, scoping down the public code API
- Page Markdown is now rendered using HtmlStrings (`{{ $content }}` instead of `{!! $markdown !!}`)
- The MarkdownDocument class is now Htmlable

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -95,8 +95,12 @@ return [
     'aliases' => [
         'Hyde' => Hyde\Hyde::class,
         'Site' => \Hyde\Facades\Site::class,
+        'Meta' => \Hyde\Facades\Meta::class,
         'Route' => \Hyde\Facades\Route::class,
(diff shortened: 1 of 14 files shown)


caendesilva:
This is why I am very happy we have Percy installed. It seems like this has broken parts of the custom Markdown system. Some Markdown parts are not being compiled, it seems to be the custom colored blockquotes and tables, so I think somewhere we lost passing data to the Markdown compiler about the rendered page so we're using the simple Markdown converter instead of the dynamic one.

PR #658, opened 2022-11-09
Overhaul file models

caendesilva:
Very much work in progress/experimental at this stage.

Experiments with adding strong path types, which in addition to being able to type hint a path type (rel/abs) it could actually also enforce/normalize these types.


packages/framework/src/Support/Filesystem/RelativePathString.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Hyde\Support\Filesystem;
+
+use Hyde\Hyde;
+use JetBrains\PhpStorm\ArrayShape;
+
+/**
+ * Denotes and enforces a path relative to the project root.
+ */
+final class RelativePathString extends PathString
```
caendesilva, line 13: Since all paths used by Hyde are (should) be relative up until the time they are used to actually interact with the file system should be relative. We only convert it to an absolute Hyde::path() before loading/saving a file. The idea behind the path-string type system is to ensure we relative paths, but that might better be onforced by only having an absolute path string type, as everything else is expected to be relative. That also better conveys the difference I think.

caendesilva:
I think we'll ditch the path string experiments as we now have https://github.com/hydephp/develop/pull/695. Multiple filemodels could still be of interest, maybe. Not sure if mimetypes needs an enum.

PR #660, opened 2022-11-11
Add an easy configuration option to store all source directories in a subfolder

caendesilva:
Changes:
- Makes the source roots configurable, for example to have all the _source files in a subdirectory named content/ (so content/_pages/)
- Adds a HydePage::path() helper to get an absolute file path to the page's source directory, or a file within it.

Todos:

- [x] Add method annotations to the Hyde facade for the added methods (getSourceRoot, etc)
- [x] https://github.com/hydephp/develop/issues/662

Code review:

- [ ] Is source_root the best name for this feature?
- [ ] Should the _media file be put in the custom folder too? (Maybe: holding off for now)
- [x] Test site can be built with custom sources (Done: tested with all page types, works perfectly)
- [x] Test the realtime compiler can handle custom source roots (Done: tested with all page types, works perfectly)


caendesilva:
Okay, so this was so much easier to do than I expected. Of course everything is probably going to break the second after I post this, but I'm feeling optimistic. Pretty much all it took was adding a single concatenation to apply the configured changes to all source files. That's why I'm running everything through the HydeKernel! Much easier to apply configurations when it all is stored in one canonical place with clearly defined entry points.

caendesilva:
Searching for files/classes that might not be compatible with this:

- PublishesHydeViews hardcodes the _pages destination
- ValidationService also hardcodes _pages

Also noted that registerOutputDirectories already unslashes paths so we don't need to do that in the HSP

Maybe a HydePage::path() helper could be useful (especially for publishing files)?

PR #668, opened 2022-11-17
Create a simple UI kit

caendesilva:
## Abstract

Adds a simple UI kit intended to making developing Hyde Blade frontends much faster by providing unopinionated layouts and helpers. This these are extracted to a new package, `hyde/ui-kit`. The views are in the components directory so the newer x-style components can be used. 

## Components - [Live Preview](https://git.desilva.se/hyde-ui-kit-demo/)

components-composite

### Focus

image-min.png)

### Sidegrid

image-min.png)


caendesilva:
I want to investigate the effect this has on the app.css filesize, an increase there could be cause for having it be separate.

caendesilva:
Adding buttons, it would be nice to dynamically use these as anchors if an href is set. Gonna take a look at that in a bit, writing here so I don't forget.

caendesilva:
I also created a code component but it's super hacky so won't include it, but adding it here for reference:

Usage:
```blade
<x-hyde::ui.components.code lang="blade">
<@x-hyde::ui.components.heading align="left">
Heading Component
<@x-hyde::ui.components.heading>
</x-hyde::ui.components.code>
```

Component:
```blade
@props(['lang' => 'txt'])
@php($slot = (str_replace('<@x', '<x', $slot)))
<div class="prose dark:prose-invert torchlight-enabled">
    {{ Hyde::markdown("```$lang\n$slot\n```", false, \Hyde\Pages\MarkdownPage::class) }}
</div>
```

caendesilva:
When creating this demo page, something struck me that could be nice, a config option to disable the footer and navigation. Perfect for one-off pages where you manually copy the HTML somewhere else like I did.

caendesilva:
Just confirmed, they pass the Lighthouse accessibility audit.

caendesilva:
As I like to keep dependencies light I'm not sure if we should include this with the base install. Having it as a separate package makes it dead easy to add on by just requiring the Composer package.

PR #675, opened 2022-11-18
Refactor publications to integrate better with Hyde

caendesilva:
Refactors the great work @rgasch has done so the changes better integrate within the Hyde core.

Notably, the data has been refactored into a few classes: both the PublicationType class for a stronger OOP experience, and the pages into "real" Hyde page classes that get auto-discovered and fully integrated into Hyde and its build process.

This also means that the build:publications command has been removed as the standard build command now compiles publications as well. The rendering logic in the removed command has been merged into the new individual page classes.

In addition, this change means that the publications can show up in navigation menus, and be linked to using routes. Try running route list command!

```
$ php hyde route:list

+---------------------+------------------------+-----------------------+------------+
| Page Type           | Source File            | Output File           | Route Key  |
+---------------------+------------------------+-----------------------+------------+
| BladePage           | _pages/404.blade.php   | _site/404.html        | 404        |
| BladePage           | _pages/index.blade.php | _site/index.html      | index      |
| PublicationPage     | blah/abc.md            | _site/blah/abc.html   | blah/abc   |
| PublicationPage     | blah/123.md            | _site/blah/123.html   | blah/123   |
| PublicationPage     | blah/aaa.md            | _site/blah/aaa.html   | blah/aaa   |
| PublicationListPage | blah/schema.json       | _site/blah/index.html | blah/index |
+---------------------+------------------------+-----------------------+------------+
```


caendesilva:
Had to delete the blah test files for now as a lot of unrelated tests fail because of that. Priority now is to get the relevant tests to pass. Likley we will readd it but using more proper data.

caendesilva:
Todo: I think publication pages should be hidden from navigation, but a navigation item linking to the listing page should be added, using the directory name as the label.

caendesilva:
I think we should update the schema to not include .blade.php as Laravel doesn't recognize view keys that includes those. Right now I'm stripping out the file extension so it still works, but if we remove the suffix from the schema we can also ship stock views like hyde::layouts.post  for example.

PR #688, opened 2022-11-24
Refactor publication commands and create actions

caendesilva:
Refactors for better usage of built-in helpers and reduces redundant code. Also improves test coverage.

### DIFF
--- a/packages/framework/src/Console/Commands/MakePublicationCommand.php
+++ b/packages/framework/src/Console/Commands/MakePublicationCommand.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Hyde\Console\Commands;
 
-use Exception;
 use Hyde\Console\Commands\Interfaces\CommandHandleInterface;
 use Hyde\Console\Concerns\ValidatingCommand;
 use Hyde\Framework\Actions\CreatesNewPublicationFile;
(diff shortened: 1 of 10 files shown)


packages/framework/src/Console/Commands/MakePublicationCommand.php
```diff
@@ -32,54 +38,45 @@ public function handle(): int
...
-                $this->output->writeln('<bg=magenta;fg=white>Exiting without overwriting existing publication file!</>');
+            } catch (FileConflictException) {
+                $this->error('Error: A publication already exists with the same canonical field value');
+                $overwrite = $this->askWithValidation(
+                    'overwrite',
+                    'Do you wish to overwrite the existing file (y/n)',
+                    ['required', 'string', 'in:y,n'],
+                    'n'
+                );
+                if (strtolower($overwrite) == 'y') {
```
caendesilva, line 67: I think we should be able to use the built in confirm helper for this

packages/framework/tests/Feature/Commands/MakePublicationCommandTest.php
```diff
@@ -31,45 +33,93 @@ protected function tearDown(): void
...
+            ->expectsOutput('Creating a new publication of type [test-publication]')
+            ->expectsQuestion('Title', 'Hello World')
             ->expectsOutput('Saving publication data to [test-publication/hello-world.md]')
             ->expectsOutput('Publication created successfully!')
             ->assertExitCode(0);
 
         $this->assertTrue(File::exists(Hyde::path('test-publication/hello-world.md')));
         $this->assertEqualsIgnoringLineEndingType('---
-__createdAt: '.Carbon::now()->format('Y-m-d H:i:s').'
+__createdAt: 2022-01-01 00:00:00
```
caendesilva, line 49: Strings like these should probably be escaped

PR #710, opened 2022-11-27
Refactor and add more baseline page unit tests

caendesilva:
Adds more tests, see https://github.com/hydephp/develop/pull/705


packages/framework/tests/Unit/Pages/BaseHydePageUnitTestMethods.php
```diff
@@ -0,0 +1,68 @@
...
+
+    public function testSourcePath();
+
+    public function testFiles();
+
+    public function testNavigationMenuLabel();
+
+    public function testGetOutputPath();
+
+    public function testGet();
```
caendesilva, line 21: Update test for https://github.com/hydephp/develop/pull/713 and add test for https://github.com/hydephp/develop/pull/712

caendesilva:
We might also want a dynamic method to ensure we test all the methods, instead of relying on the static interface being up to date.

caendesilva:
We should also add tests for the methods added by the BaseMarkdownPage class

PR #716, opened 2022-11-28
Committing code to 'tag' field type.

rgasch:
Mainly it includes the new tag feature, but also some other minor improvements such as: 

1) Use trim() on input where appropriate
2) Changed min/max types to string, this makes sense because it could be a datetime or A-Z or something like that
3) Simplified the logic for MAX_RETRIES on the validation  checker for input, I think this version reads nicer than the original one (sidenote: this method doesn't really need to be recursive, it could also be done in a do...while loop)
4) Some other minor things

Please look for the string FIXME in the code, I used this to mark some things which I think you should review just to make sure. It should not be anything to fix, but just to review.

The new command is called makePublicationTag ... I called it "tag" because this is the most obvious use case, but it's basically the action of chosing an item from a list, doesn't have to be tag
This command creates (or adds to) a file called tags.json ... the fact that we now have another file in the root directory IMHO strengthens the case for the argument that all content files should be under a content directory

Also, fom the above paragraph you will deduce that tags are global. Yes they are ... initially I had a version which did both global and per-publication tags but the code got messy and there are some edge cases which appear which you deal with this configuration ... so the global version is much simpler (code wise) so this is what I opted for. I think this is OK because for most applications, you wont have like 50 tags, probably more like 5-10 max so this is OK
Give it a try and let me know what you think

Oh yes, I also added a check to make sure you can't add the same filename twice when defining a publication type

### DIFF
--- a/packages/framework/src/Console/Commands/MakePublicationTagCommand.php
+++ b/packages/framework/src/Console/Commands/MakePublicationTagCommand.php
@@ -1,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Hyde\Console\Commands;
+
+use Hyde\Console\Commands\Interfaces\CommandHandleInterface;
+use Hyde\Console\Concerns\ValidatingCommand;
+use Hyde\Framework\Features\Publications\PublicationService;
+use Hyde\Hyde;
+use Illuminate\Support\Str;
+use LaravelZero\Framework\Commands\Command;
+use function Safe\json_decode;
+use function Safe\json_encode;
+use function Safe\file_put_contents;
+
+/**
+ * Hyde Command to create a new publication type.
+ *
+ * @see \Hyde\Framework\Testing\Feature\Commands\MakePublicationTypeCommandTest
+ */
+class MakePublicationTagCommand extends ValidatingCommand implements CommandHandleInterface
+{
+    /** @var string */
+    protected $signature = 'make:publicationTag';
+
+    /** @var string */
+    protected $description = 'Create a new publication type tag definition';
+
+    public function handle(): int
+    {
+        $this->title('Creating a new Publication Type Tag!');
+
+        $filename = Hyde::pathToRelative('tags.json');
+        $tags     = PublicationService::getAllTags();
+        $tagName  = $this->askWithValidation('name', 'Tag name', ['required', 'string']);
+        if (isset($tags[$tagName])) {
+            $this->output->error("Tag [$tagName] already exists");
+            return Command::FAILURE;
+        }
+
+        $lines = [];
+        $this->output->writeln('<bg=magenta;fg=white>Enter the tag values (end with an empty line):</>');
+        do {
+            $line = Str::replace("\n", '', fgets(STDIN));
+            if ($line === '') {
+                break;
+            }
+            $lines[] = trim($line);
+        } while (true);
+        $tags[$tagName] = $lines;
+
+        $this->output->writeln(sprintf('Saving tag data to [%s]', $filename));
+        file_put_contents($filename, json_encode($tags, JSON_PRETTY_PRINT));
+
+        return Command::SUCCESS;
+    }
+}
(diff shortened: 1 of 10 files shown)


packages/framework/src/Console/Commands/MakePublicationTypeCommand.php
```diff
@@ -109,41 +112,69 @@ protected function captureFieldsDefinitions(): Collection
...
+            $this->line('  10 - Tag (select value from list)');
+            $type = (int) $this->askWithValidation('type', 'Field type (1-10)', ['required', 'integer', 'between:1,10'], 1);
+
+            if ($type < 10) {
+                do {
+                    // TODO This should only be done for types that can have length restrictions right?
+                    // ANSWER: No, it can also be a value restriction
+                    // - (int: 0 - 2022)
+                    // - (float: 0 - 360)
+                    // - (datetime: 2022-01-01 - 2022-12-31)
```
caendesilva, line 142: Oh yeah right, thanks! Question: are there any types that can't have length restrictions? How about images?

packages/framework/src/Console/Commands/MakePublicationTagCommand.php
```diff
@@ -0,0 +1,58 @@
...
+        $filename = Hyde::pathToRelative('tags.json');
+        $tags     = PublicationService::getAllTags();
+        $tagName  = $this->askWithValidation('name', 'Tag name', ['required', 'string']);
+        if (isset($tags[$tagName])) {
+            $this->output->error("Tag [$tagName] already exists");
+            return Command::FAILURE;
+        }
+
+        $lines = [];
+        $this->output->writeln('<bg=magenta;fg=white>Enter the tag values (end with an empty line):</>');
```
caendesilva, line 43: image

It's not terminating for me, I think it might be because I'm on Windows and for some reason I truly don't understand Windows uses "\r\n" instead of "\n" for terminating newlines.

caendesilva:
> 1. Use trim() on input where appropriate

Sounds good.

> 2. Changed min/max types to string, this makes sense because it could be a datetime or A-Z or something like that

Okay yeah this is definitively a good idea since the input captured is also strings. We can then cast actual integers to integers when/if we need it.

> 3. Simplified the logic for MAX_RETRIES on the validation  checker for input, I think this version reads nicer than the original one (sidenote: this method doesn't really need to be recursive, it could also be done in a do...while loop)

I'll take a look. The reason I added this was that when testing if the answer wasn't correct it'd get stuck in an infinite loop. Same if you ran the command with the "-n" flag.

> Please look for the string FIXME in the code, I used this to mark some things which I think you should review just to make sure. It should not be anything to fix, but just to review.

Will do. I recently added a static analyser that will fail + add comments when a fixme is found. 

> The new command is called makePublicationTag ... I called it "tag" because this is the most obvious use case

Tag seems appropriate.

> the fact that we now have another file in the root directory IMHO strengthens the case for the argument that all content files should be under a content directory
I'm very inclined to agree.

> I think this is OK because for most applications, you wont have like 50 tags, probably more like 5-10 max so this is OK Give it a try and let me know what you think

That makes sense.

> Oh yes, I also added a check to make sure you can't add the same filename twice when defining a publication type

I'm not seeing this in the diff. Where is it?

packages/framework/src/Framework/Features/Publications/Models/PublicationFieldType.php
```diff
@@ -29,27 +29,29 @@ class PublicationFieldType implements SerializableContract
     ];
 
     public readonly string $type;
-    public readonly ?int $max;
-    public readonly ?int $min;
+    public readonly string $max; // FIXME: needs to be string (dates, etc), should not be null, default = 0?
```
caendesilva, line 32: Yeah, using strings definitely make sense. As for not being null, what if a user doesn't want length/range validation what would be the default value? `0` is a valid range, I had null to designate that the setting is not applied. What do you think makes most sense for this case?

packages/framework/src/Console/Commands/MakePublicationTypeCommand.php
rgasch: Currently, every field except tags asks for a length ... this should/could be refined (I'll take care of it). 

The only problem is that it's context-dependent. For strings, it's the string length, for numbers it's the numeric value, for dates it's the date range (which isn't validated properly yet), so let me take a look and see how to best do this.

caendesilva: Okay, sounds great. Thanks!

caendesilva (commented):
Just gonna resolve the todos then I'll merge!

caendesilva (approved):

PR #721, opened 2022-12-02
Add publications seeder and field type validation 

rgasch:


caendesilva:
I've done some refactors, lemme know if there are any merge conflicts or other issues. I can help take care of them if you want.

caendesilva:
@rgasch Okay updated your branch and tried to resolve the merge conflicts, starting to think maybe we shouldn't work on the same files at the same time 😅

packages/framework/src/Framework/Features/Publications/PublicationService.php
```diff
@@ -51,29 +51,40 @@ public static function getPublicationsForPubType(PublicationType $pubType): Coll
...
-    public static function getAllTags(): Collection
+    public static function getAllTags(bool $reload=true): Collection
     {
         $filename = Hyde::pathToRelative('tags.json');
         if (! file_exists($filename)) {
             return Collection::create();
         }
 
-        return Collection::create(json_decode(file_get_contents($filename), true))->sortKeys();
+        static $tags = null;
```
caendesilva, line 82: What's the purpose of the static variable here? Is it just to cache the discovered tags? Or does this actually make so the data is reloaded automatically when using the returned variable?

I haven't really used (non-class) static variables before.

PR #737, opened 2022-12-07
Integrate publications seeder 

caendesilva:
Integrates code from https://github.com/hydephp/develop/pull/721

### DIFF
--- a/packages/framework/src/Framework/Actions/SeedsPublicationFiles.php
+++ b/packages/framework/src/Framework/Actions/SeedsPublicationFiles.php
@@ -96,15 +96,15 @@ class SeedsPublicationFiles extends CreateAction implements CreateActionContract
     {
         return match ($field->type->value) {
             'array' => $this->getArrayItems(),
-             'boolean' => rand(0, 100) < 50,
-             'datetime' => $this->getDateTimeValue(),
-             'float' => rand(-10000000, 10000000) / 100,
-             'image' => 'https://picsum.photos/id/'.rand(1, 1000).'/400/400',
-             'integer' => rand(-100000, 100000),
-             'string' => substr($this->fakeSentence(10), 0, rand(0, 255)),
-             'tag' => $this->getTags($field),
-             'text' => $this->getTextValue(rand(3, 20)),
-             'url' => $this->fakeUrl(),
+            'boolean' => rand(0, 100) < 50,
+            'datetime' => $this->getDateTimeValue(),
+            'float' => rand(-10000000, 10000000) / 100,
+            'image' => 'https://picsum.photos/id/'.rand(1, 1000).'/400/400',
+            'integer' => rand(-100000, 100000),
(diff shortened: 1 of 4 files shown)


packages/framework/tests/Feature/Actions/SeedsPublicationFilesTest.php
```diff
@@ -0,0 +1,134 @@
...
+
+        return $files;
+    }
+
+    protected function firstPublicationFilePath(): string
+    {
+        return $this->getPublicationFiles()[0];
+    }
+
+    protected function assertFileMatchesString(string $expected, string $filepath, ?int $limit = null)
```
caendesilva, line 89: It might actually be easier to just parse the file as a simple markdown document object and check the matter through that.

caendesilva:
@rgasch what do you think about adding a seeder for publication types too?

rgasch:
> Generating 1000 publications took 847ms (0.85ms/each) on my Windows machine which has a comparably slow filesystem. I'm very impressed @rgasch!

Thank you!

rgasch:
> @rgasch what do you think about adding a seeder for publication types too?

I'm not sure how that would work. Because to be able to see publication types, you need to know what fields those pubtypes would contain (and since publication types need a context in order to make contextual sense, just generating semi-random pubtypes would not be very clear) 
+ since the creation of a publication type is essentially guided (through the CLI questions interface) I don't think it's that necessary to have a seeder. 

Also, if we deliver a nice and complex/complete publication type as a demo/sample, I think the usage of publication types should be pretty self expanatory. 

The reason why I wrote the seeder for pulications was to 
1) Enable to you quickly generate content to test and develop your theme/layout
2) Performance testing

packages/framework/src/Framework/Actions/SeedsPublicationFiles.php
```diff
@@ -0,0 +1,137 @@
...
+                    $canonicalValue = $field->name == $canonicalFieldName ? $value : '';
+                    break;
+                case 'string':
+                    $min = $field->min ?: 0;
+                    $max = $field->max ?: 255;
+                    $value = Str::of($faker->sentence(10))->limit($faker->numberBetween($min, $max), '...');
+                    $output .= "$field->name: $value\n";
+                    $canonicalValue = $field->name == $canonicalFieldName ? $value : '';
+                    break;
+                case 'tag':
```
caendesilva, line 111: I'm still not able to get the tags to work, could you explain what they're supposed to do @rgasch?

rgasch: I will take a look tonight ...

caendesilva:
Since I think it's wasteful to add a whole dependency for such a comparatively small feature using so little code, I created a simple faker alternative in https://github.com/hydephp/develop/commit/e99b3d46e3eb76d168b224bca7fc58fcd9ced215.

PR #739, opened 2022-12-08
Get page routes from the route index

caendesilva:
I think there's no point in creating a bunch of new objects each time we want to access a route, however since only pages with actual source files generally get added to the route index, the old behaviour is present as a fall back (for example when using `new MarkdownPage` in a test)

### DIFF
--- a/monorepo/HydeStan/HydeStan.php
+++ b/monorepo/HydeStan/HydeStan.php
@@ -31,8 +31,10 @@ class HydeStan
             number_format(count($this->files)))
         );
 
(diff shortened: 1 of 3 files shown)


caendesilva:
This causes an issue with multiple source files with the same identifier, this is not an issue in actual code as you can't have multiple files with the same name anyways, but in tests where we do everything in-memory it can be an issue as seen here:
https://github.com/hydephp/develop/blob/a41fa35042fe9fce2880e3701b972fe990b52be5/packages/framework/tests/Feature/HydePageTest.php#L958-L962

Issue #669, opened 2022-11-17
Make navigation behaviour for nested pages configurable


caendesilva:
Say we have a structure like this, where pages are in a directory of their source directory:

```
_pages/about/foo.md
_pages/about/bar.md
```

How should Hyde handle the navigation for these pages? Currently they get added to the navigation menu, but maybe they shouldn't? I have a few ideas for improvement, and I think this should be configurable. I am very happy to get input on this, and also which option should be the default.

1. Add the pages to the navigation menu (current behaviour)
2. Put them in an automatic dropdown based on the subdirectory name (my fav)
3. Don't show them at all

PR #740, opened 2022-12-08
Make navigation behaviour for nested pages configurable

caendesilva:
Implements the data generation backend for this feature.  We are still going to need to actually implement frontend code to support automatic dropdowns.

> https://github.com/hydephp/develop/issues/669
>
> Say we have a structure like this, where pages are in a directory of their source directory:
> 
> ```
> _pages/about/foo.md
> _pages/about/bar.md
> ```
> 
> How should Hyde handle the navigation for these pages? Currently they get added to the navigation menu, but maybe they shouldn't? I have a few ideas for improvement, and I think this should be configurable. I am very happy to get input on this, and also which option should be the default.
> 
> 1. Add the pages to the navigation menu (current behaviour)
> 2. Put them in an automatic dropdown based on the subdirectory name (my fav)
> 3. Don't show them at all

### DIFF
--- a/config/hyde.php
+++ b/config/hyde.php
@@ -177,6 +177,10 @@ return [
         'custom' => [
             // NavItem::toLink('https://github.com/hydephp/hyde', 'GitHub', 200),
         ],
+
+        // How should pages in subdirectories be displayed in the menu?
+        // You can choose between 'dropdown', 'flat', and 'hidden'.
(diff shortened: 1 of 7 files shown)


caendesilva:
We are still going to need to actually implement frontend code to support automatic dropdowns

PR #744, opened 2022-12-08
Turn navigation items with groups into automatic dropdowns

caendesilva:


packages/framework/src/Framework/Features/Navigation/NavigationMenu.php
```diff
@@ -80,4 +80,17 @@ protected function filterDocumentationPage(NavItem $item): bool
             && $item->route->getPage() instanceof DocumentationPage
             && $item->route->getRouteKey() !== 'docs/index';
     }
+
+    public function hasDropdowns(): bool
```
caendesilva, line 84: Since these methods can't be used by the documentation sidebar that extends this class, maybe we should split this out to a dedicated class for the "main" navigation so to speak.

caendesilva: Moved to https://github.com/hydephp/develop/issues/745

PR #746, opened 2022-12-08
Refactor navigation services

caendesilva:
Add a new base class and refactors to reduce duplicate code between navigation menu types. The main feature however is how dropdowns are handled, as they are now added as DropdownNavigationItems in the main collection array.

Contains the following breaking changes:

- The filtering of duplicate items based on the label not link
- API: Removed unused `NavigationMenu::$currentRoute` class property

### DIFF
--- a/packages/framework/src/Framework/Features/Navigation/NavigationMenu.php
+++ b/packages/framework/src/Framework/Features/Navigation/NavigationMenu.php
@@ -70,14 +70,16 @@ class NavigationMenu extends BaseNavigationMenu
         return count($this->getDropdowns()) >= 1;
     }
 
-    /** @return array<string, array<NavItem>> */
+    /** @return array<string, DropdownNavItem> */
     public function getDropdowns(): array
     {
         if (! $this->dropdownsEnabled()) {
             throw new BadMethodCallException('Dropdowns are not enabled. Enable it by setting `hyde.navigation.subdirectories` to `dropdown`.');
         }
 
-        return $this->dropdowns;
+        return $this->items->filter(function (NavItem $item): bool {
+            return $item instanceof DropdownNavItem;
+        })->all();
     }
 
     protected static function canBeInDropdown(NavItem $item): bool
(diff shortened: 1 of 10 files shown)


packages/framework/src/Framework/Features/Navigation/NavigationMenu.php
```diff
@@ -19,9 +19,7 @@
  */
 class NavigationMenu
 {
-    public Route $currentRoute;
-
-    public Collection $items;
+    protected Collection $items;
```
caendesilva, line 22: Actually, since the `$items` property is a collection that is itself writeable, making it protected makes no effective change as you can still modify it using the getter.

caendesilva: We could handle this by using the getter to create a new collection instance in case we do want to scope down the API and make it read-only (you can still push pages to the router). We'll see...

caendesilva:
Commit e7412b41e7 breaks navigation menu ordering as the numerical keys are what determines the sort order. I removed this so we could use `put` instead of `push` so we later can `forget` items that are moved to dropdowns. We'll have to use filtering instead.

caendesilva:
> Commit [e7412b4](https://github.com/hydephp/develop/commit/e7412b41e78a1e04cc400833c339e0b6e21be1fa) breaks navigation menu ordering as the numerical keys are what determines the sort order. I removed this so we could use `put` instead of `push` so we later can `forget` items that are moved to dropdowns. We'll have to use filtering instead.

I find it much easier to interact with the underlying array when it's associative. Maybe we can reindex it according to sorting order when calling the sort method? (Since that should be the last data manipulation step)

caendesilva:
> > Commit [e7412b4](https://github.com/hydephp/develop/commit/e7412b41e78a1e04cc400833c339e0b6e21be1fa) breaks navigation menu ordering as the numerical keys are what determines the sort order. I removed this so we could use `put` instead of `push` so we later can `forget` items that are moved to dropdowns. We'll have to use filtering instead.
> 
> I find it much easier to interact with the underlying array when it's associative. Maybe we can reindex it according to sorting order when calling the sort method? (Since that should be the last data manipulation step)

Oh, right, we're already doing this...

PR #751, opened 2022-12-11
Refactor the navigation data factory

emmadesilva:
Refactors the factory to be more readable, and adds more paths for a more forgiving front matter structure.


caendesilva:
Benchmark just for fun. After change: `0.040ms`, before: `0.033ms`. 10 000 iterations. 7 microseconds difference. That's just under the time taken by light to travel one mile in typical single-mode fiber optic cable. I'm okay with that.

PR #760, opened 2022-12-16
Refactor the publication file creator

caendesilva:


packages/framework/src/Framework/Actions/CreatesNewPublicationPage.php
```diff
@@ -79,6 +79,7 @@ protected function handleCreate(): void
     protected function getCanonicalValue(PublicationFieldType $canonicalFieldDefinition, string $canonicalFieldName): string
     {
         try {
+            // TODO: Is it reasonable to use arrays as canonical field values?
```
caendesilva, line 81: Is it reasonable to use arrays as canonical field values?

PR #761, opened 2022-12-16
Internal: Refactor testing helpers

caendesilva:


caendesilva:
I think in general we should try to normalize all data input to use Unix endings, so most of the latest commits are about that.

PR #764, opened 2022-12-18
Support meta fields as canonical publication fields

caendesilva:


caendesilva:
Continuing work in https://github.com/hydephp/develop/pull/760

PR #765, opened 2022-12-19
Add tests for and refactor publication field type validation rules

caendesilva:


packages/framework/src/Framework/Features/Publications/Models/PublicationFieldType.php
```diff
@@ -64,25 +65,26 @@ public function toArray(): array
...
         if (($this->min && ! $this->max) || ($this->min == '0' && $this->max == '0')) {
             $fieldRules->forget($fieldRules->search('required'));
-            $doBetween = false;
+            $useRange = false;
         }
 
         switch ($this->type->value) {
             case 'array':
-                $fieldRules->add('array');
+                $fieldRules->add('array'); // FIXME do we do range validation too?
```
caendesilva, line 84: Should we do range validation on arrays too?

packages/framework/src/Framework/Features/Publications/Models/PublicationFieldType.php
```diff
@@ -91,26 +93,26 @@ public function getValidationRules(bool $reload = true): Collection
...
-                if ($doBetween) {
+                if ($useRange) {
                     $fieldRules->add("between:$this->min,$this->max");
                 }
                 break;
             case 'image':
                 $mediaFiles = PublicationService::getMediaForPubType($this->publicationType, $reload);
                 $valueList = $mediaFiles->implode(',');
-                $fieldRules->add("in:$valueList");
+                $fieldRules->add("in:$valueList"); // FIXME What if the list is empty?
```
caendesilva, line 103: A few questions here:

1. What if the list is empty?
2. Now the items look like 'in:_media/foo/bar.jpg', but do we really need the directory information? Wouldn't it suffice with just 'in:bar.jpg' since we already know what directory it is in? We could then easily qualify it within the template and/or via a helper method.

packages/framework/src/Framework/Features/Publications/Models/PublicationFieldType.php
```diff
@@ -91,26 +93,26 @@ public function getValidationRules(bool $reload = true): Collection
...
+                                                   //   We could then easily qualify it within the template and/or via a helper method.
                 break;
             case 'tag':
-                $tagValues = PublicationService::getValuesForTagName($this->tagGroup, $reload);
+                $tagValues = PublicationService::getValuesForTagName($this->tagGroup, $reload) ?? collect([]);
                 $valueList = $tagValues->implode(',');
                 $fieldRules->add("in:$valueList");
                 break;
             case 'url':
+                // FIXME Shouldn't we add a 'url' rule here?
```
caendesilva, line 114: Shouldn't we add a 'url' rule here?

packages/framework/src/Framework/Features/Publications/Models/PublicationFieldType.php
```diff
@@ -91,26 +93,26 @@ public function getValidationRules(bool $reload = true): Collection
...
                 break;
-            default:
```
caendesilva, line 110: Default case removed as it's not possible to reach since the backed enum is evaluated in the constructor.

caendesilva (commented):
@rgasch Whenever you have time over, please take a look at my comments and questions.

packages/framework/src/Framework/Features/Publications/Models/PublicationFieldType.php
rgasch: Well, this has the potential to open a can of works. Because what do we mean? The # if items in the array? The lengths of the individual items in the array? Both? I'd rather just let the user enter whatever, because we could just be adding unnecessary complexity.

caendesilva: I was thinking it would limit the number of entries in the array. But you're right, it likely adds more complexity than value.

packages/framework/src/Framework/Features/Publications/Models/PublicationFieldType.php
caendesilva: Okay so for the first one, I think it should be left blank as I assume that allows nothing to be selected, which should be the intended behaviour when there is nothing to select from.

For the second, I think the principle of least astonishment here resolves to keep it as it is now for clarity. At least for now.

caendesilva:
The range validations add a very large amount of complexity. I'm starting to think about how much we really need them. Sure length validation for article descriptions might be good, but do we really need to validate date ranges and integer values? Maybe it would be better to allow arbitrary Laravel validation rules to be specified in the schema file instead of its min/max values.

PR #774, opened 2022-12-24
Refactor the make:publication command class 

caendesilva:


caendesilva:
In the last PR I made so that some types could not be selected as canonical when using the make pubType command, however since files can be edited arbitrary, other types can be set manually. I worked around this in https://github.com/hydephp/develop/pull/774/commits/38d02b5617a51a134b3fc461ccdb0963b9a1f732

PR #778, opened 2022-12-27
Refactor view finder for publication templates to check file extensions instead of if view exists

caendesilva:
Refactors to check if template key ends with .blade.php and using that to signify if it's an anonymous view. This leads to a more verbose and intuitive system. If you specify a filename, Hyde will use the filename. If not, Hyde will assume it's a globally registered view.

This may break code on the [publications-feature](https://github.com/hydephp/develop/tree/publications-feature) branch as the schemas need to be updated: (see https://github.com/hydephp/develop/commit/9b471efa09c8c8ecd9b9d8ce394f0e5bb26daca9 for an example)

```diff
- "detailTemplate": "detail",
- "listTemplate": "list",
+ "detailTemplate": "detail.blade.php",
+ "listTemplate": "list.blade.php",
```

The files are still assumed to be in the same directory as the schema.json file.

### DIFF
--- a/packages/framework/src/Framework/Actions/CreatesNewPublicationType.php
+++ b/packages/framework/src/Framework/Actions/CreatesNewPublicationType.php
@@ -87,7 +87,7 @@ class CreatesNewPublicationType extends CreateAction implements CreateActionCont
         @endsection
         BLADE;
 
-        $this->savePublicationFile("detail.blade.php", $contents);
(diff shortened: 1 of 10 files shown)


caendesilva:
This diff https://github.com/hydephp/develop/commit/9b471efa09c8c8ecd9b9d8ce394f0e5bb26daca9 might be useful for finding files for normalizing all these test schemas to use the same names

PR #685, opened 2022-11-23
Add a package that provides a publications feature

caendesilva:
## About

Publication pages adds an easy way to create custom no-code page types, with support using a custom front matter schema and Blade templates. This feature is in development at the HydePHP develop monorepository, https://github.com/hydephp/develop/pull/685, and will be published as an official HydePHP extension tied to HydePHP v1.0.

Thank you to @rgasch for suggesting and implementing this feature in https://github.com/hydephp/develop/pull/667!

### General

- [ ] Add documentation for the publications feature (@rgasch do you want to handle this?)
- [x] Ensure 100% test coverage for publications code
- [x] Resolve [code review comments](https://github.com/hydephp/develop/pull/685/files#conversations-menu) (20)

### Pull Requests
- [x] https://github.com/hydephp/develop/pull/729
- [x] https://github.com/hydephp/develop/pull/814
- [x] https://github.com/hydephp/develop/pull/839 (related: Schema option?)
- [x] https://github.com/hydephp/develop/pull/837
- [x] https://github.com/hydephp/develop/pull/841
- [x] https://github.com/hydephp/develop/pull/840

### Issues
- [x] https://github.com/hydephp/publications/issues/2
- [x] https://github.com/hydephp/publications/issues/3
- [x] https://github.com/hydephp/publications/issues/5
- [x] https://github.com/hydephp/publications/issues/4

### Comments

- [x] Generate list and detail pages when creating publication types ([comment](https://github.com/hydephp/develop/pull/685#issuecomment-1328113544))
- [x] Generated pubtype field names should be lower/kebab case ([comment](https://github.com/hydephp/develop/pull/685#issuecomment-1328339860))
- [x] Move logic to SourceFileParser class ([comment](https://github.com/hydephp/develop/pull/685#discussion_r1033004814))
- [x] Create class for the tag collection ([comment](https://github.com/hydephp/develop/pull/685#issuecomment-1336199055))

### Ideas/Bugs/Other

- [x] Add visual regression tests for generated pages? (skipping for now)
- [x] Replace validation message placeholders like `validation.required`
- [x] Maybe we want to validate the schema files when running the validate command?
- [x] Store __createdAt publications field as string instead of Unix timestamp (unable to repro atm)
- [x] Normalize command names

#### Updates to the make:publication command #786

- [x] When using `php hyde make:publication foo` and there are no publications at all, the current message `"Error: Unable to locate any publication types. Did you create any?"` might not be entirely relevant.
- [x] Fix integers are serialized as strings
[...]

PR #786, opened 2022-12-28
Update the make:publication command

caendesilva:
Refactors the `make:publication` command to fix bugs and provide a consistent console UI layer.

Fixes:

- [x] Don't ask users to fill in meta fields
- [x] Fix bug causing booleans to not be fillable (presumably as writing `true` to `STDIN` is the same as writing `'true'`)
- [x] Indent empty collection warnings
- [x] Normalize output styles for messages to enter array and text input
- [x] Add message when skipping a field?
- [x] Test that fields that have a required rule cannot be skipped
- [x] Unset skipped values when generating front matter?
- [x] Check if we should use `$field->getValidationRules()` instead of `$field->type->rules()` as it seems we might be missing checking custom rules
- [ ] When using `php hyde make:publication foo` and there are no publications at all, the current message `"Error: Unable to locate any publication types. Did you create any?"` might not be entirely relevant. 
- [ ] Fix integers are serialized as strings
- [ ] Make sure that the proper type is selected when asked

### DIFF
--- a/packages/framework/src/Console/Commands/Helpers/InputStreamHandler.php
+++ b/packages/framework/src/Console/Commands/Helpers/InputStreamHandler.php
@@ -56,8 +56,8 @@ class InputStreamHandler extends InvokableAction
     }
 
     /** Format a consistent message that can be output to the console */
-    public static function formatMessage(string $name): string
+    public static function formatMessage(string $name, string $type = 'values'): string
     {
-        return "Enter $name (end with an empty line)";
+        return "<info>Enter $type for field </>[<comment>$name</comment>] (end with an empty line)";
     }
 }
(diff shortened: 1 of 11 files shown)


caendesilva:
>  Fix bug causing booleans to not be fillable (presumably as writing true to STDIN is the same as writing 'true')

According to the docs (https://laravel.com/docs/9.x/validation#rule-boolean)
> "The field under validation must be able to be cast as a boolean. Accepted input are true, false, 1, 0, "1", and "0"."

Just extracting bools to a helper method that casts to bool changes the generated YAML from `name: '1'` to `name: true`

Not sure how to allow strings without losing custom validation rules and without too much added code

caendesilva:
Some things left to do, moving todos to https://github.com/hydephp/develop/pull/685

Issue #54, opened 2022-06-03
Document colored blockquotes and code block filepaths


caendesilva:
Should probably be added to the advanced markdown page

caendesilva:
# Markdown

> Normal Blockquote

>info Info Blockquote

>warning Warning Blockquote

>danger Danger Blockquote

>success Success Blockquote


```php
// Filepath: Hello.php

// Foo
echo 'Hello World';

$example = "bar";
```


```php
// Filepath: Hello.php
// Foo
echo 'Hello World';

$example = "bar";
```

```markdown
// Filepath: _docs\bladedown.md

# Hello World
```

```yaml
# filepath test.yml

hello: world
- foo: "bar"
```


```yaml
# filepath nothingelsehere.yml
```

```markdown
 > Normal Blockquote

 >info Info Blockquote
 >warning Warning Blockquote
 >danger Danger Blockquote
 >success Success Blockquote
```



````markdown
// Filepath: _docs\markdown-features.md

# Automatic Filepaths!

You can now add a filepath to your code blocks
using semantic comments right in your code examples!

Example:
```markdown
// Filepath: _docs\markdown-features.md // HYDE! {"shortcodes": false} HYDE! // 

# Automatic Filepaths! 
[...]
```
````

PR #796, opened 2023-01-01
Document colored blockquotes and code block filepaths

caendesilva:
Fixes https://github.com/hydephp/develop/issues/54

Keeping this open for just a little bit as I'm in the car and cannot test to double check that the changes work as I'm using the GitHub web editor.


caendesilva:
Fun fact, pre and post-processing one filepath codeblock takes 0.155ms

PR #808, opened 2023-01-05
Pagination experiments 

caendesilva:
This is gonna break like all of the tests, so there is no way I'll merge the PR as it is now, I'm just playing around, later I'll cherry pick the changes I like to a new PR. I just want to get the CI's opinion on this.

PR #814, opened 2023-01-06
Add pagination feature to publications

caendesilva:
Integrates working changes from https://github.com/hydephp/develop/pull/808


packages/framework/src/Framework/Features/Publications/PaginationService.php
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * Internal developer note: It may be useful to match the Laravel method names as much as possible.
+ *
+ * @link https://laravel.com/docs/9.x/pagination#paginator-instance-methods
+ *
+ * Additional note: In most cases when this class uses the term "page" it is referring
+ * to the paginated page of publications, not the publication pages themselves.
+ * This is for the reason stated above, to match the Laravel method names.
+ */
+class PaginationService
```
caendesilva, line 23: Consider renaming class to `Paginator`

packages/framework/resources/views/layouts/publication_list.blade.php
```diff
@@ -4,13 +4,28 @@
...
             @php/** @var \Hyde\Framework\Features\Publications\Models\PublicationType $publicationType*/@endphp
             <h1>Publications for type {{ $publicationType->name }}</h1>
-            <ol>
-                @foreach($publicationType->getPublications() as $publication)
-                    <li>
-                        <x-link :href="$publication->getRoute()">{{ $publication->title }}</x-link>
-                    </li>
-                @endforeach
-            </ol>
+            @if($publicationType->usesPagination())
```
caendesilva, line 7: Probably want to extract this to a separate view to keep it light

packages/framework/src/Framework/Features/Publications/Paginator.php
```diff
@@ -0,0 +1,175 @@
...
+use Hyde\Framework\Features\Publications\Models\PaginationSettings;
+use Hyde\Hyde;
+use Illuminate\Contracts\Support\Arrayable;
+use Illuminate\Support\Collection;
+use function is_array;
+
+/**
+ * @see \Hyde\Framework\Testing\Feature\PaginationServiceTest
+ *
+ * @todo Implement sorting from PaginationSettings
```
caendesilva, line 18: Implement sorting from PaginationSettings

packages/framework/src/Framework/Features/Publications/PaginationService.php
caendesilva: Resolved in 8dbbbcc1943f1546b0093150d1128ecb9a0b99e6

caendesilva:
Since the only option from the pagination settings class we use is the pageSize we might as well inline that, and run the sorting before passing data to the paginator class, that helps keeping it more generic.

packages/framework/resources/views/layouts/publication_list.blade.php
caendesilva: Thinking a nice flow could be when creating a publication type, instead of asking if the user wants to configure pagination, we ask if the want to enable it. If enabled we present the questions (with defaults that can just be entered by hitting enter), and then we copy over the paginated template, if not enabled, we just use the default one.

caendesilva:
Thinking a nice flow could be when creating a publication type, instead of asking if the user wants to configure pagination, we ask if the want to enable it. If enabled we present the questions (with defaults that can just be entered by hitting enter), and then we copy over the paginated template, if not enabled, we just use the default one.

PR #818, opened 2023-01-08
Inline the publication settings class into the publication type class

caendesilva:
Since this class has now been simplified, and both sortField and sortAscending can be used outside of pagination, this class is now longer needed and can be inlined into the PublicationType class.


caendesilva:
We might even want to sort the collection on parse

PR #821, opened 2023-01-09
Refactor publications into the kernel container

caendesilva:


caendesilva:
Closing as this will need to be incorporated into https://github.com/hydephp/develop/pull/827

PR #822, opened 2023-01-09
Simplify the kernel boot process for a more predictable and consistent application state

caendesilva:
Right now, the Kernel uses a hybrid-dynamic booting system initiator. The Kernel can be booted both when it is needed and when called last in the HydeServiceProvider boot method. While this has been convenient so far, as the the Kernel will always be booted when calling a method that requires it, it does make it confusing and hard to know when the Kernel actually is booted. This poses a problem when exposing the Kernel to package developers to add their own integrations. For that reason I think it's best to lock down the booting process to only happen last in the HSP booter. That way package developers have until then to register their stuff. 

Todo: Update docs


caendesilva:
For context, the kernel takes about 15ms to boot on my machine.
  
```php
Benchmark::dd([function () {
    $this->files = FileCollection::boot($this);
}, function () {
    $this->pages = PageCollection::boot($this);
}, function () {
    $this->routes = RouteCollection::boot($this);
}]);

// Results:
FileCollection::boot => "4.792ms" 
PageCollection::boot => "6.221ms"
RouteCollection::boot => "0.612ms"
```

caendesilva:
Noticed a side-effect of this. In tests, if we add a file after the kernel is booted we need to reboot the kernel, this is fine most times, but now the kernel is booted before the test method is called so we can't add any files without rebooting. First I thought this was just in tests and that'd be fine, then I noticed that `PublishViewsCommand.php` needs it as that runs in the same app lifecycle. Maybe adding back the deferred booting is good, but could add a "ready to boot" state to prevent booting until after the HSP is done to provide the same strict time frame.

caendesilva:
Closed in favour of https://github.com/hydephp/develop/pull/825

PR #826, opened 2023-01-09
Extract publications package

caendesilva:
Experiments with extracting the publications feature https://github.com/hydephp/develop/pull/685 into a sub-package. The dependencies (like validation and translation are also moved to that package)

TODO

- [x] Move actions and related code to package
- [x] Move tests?


caendesilva:
Okay, that was surprisingly painless. (Famous last words)

Issue #56, opened 2022-05-16
Add some kind of pagination for blog post feeds


caendesilva:
The blog post feed on the Hyde site is getting pretty long. Time for some pagination! Any ideas?

caendesilva:
Closing, I think this is better to have as a plugin as it's a rather complex feature. If anyone comes across this and wants to implement it, or who just see the need for it, I'm open to adding it. Just comment and we can reopen the issue.

PR #832, opened 2023-01-13
Add a pagination component

caendesilva:
## About

Adds a class to aid in creating paginated pages, and with a corresponding component to render the links. The PR does not implement any actual pagination, just the framework to do so.

Cherry picks the pagination feature from https://github.com/hydephp/develop/pull/685

See also https://github.com/hydephp/develop/issues/56

### Usage

Simply create a paginator instance with a collection or array of items (like pages), and render the component. You will also likely need to find some way to pass the current page being rendered (if you're on pagination page 3, pass that to the constructor)

```blade
@php
    $paginator = new \Hyde\Framework\Features\Paginator(
        items: \Hyde\Pages\MarkdownPost::getLatestPosts(), // The items to paginate
        pageSize: 25, // How many items to show on each page
        currentPageNumber: 3, // The current page index
        paginationRouteBasename: 'posts' // Links will be 'posts/page-1.html' instead of 'page-1.html'
    )
@endphp
@include('hyde::components.pagination-navigation')
```

### Screenshots

Verbose version:

localhost_8080_ (2)
localhost_8080_ (3)

Compact version:
localhost_8080_ (5)
localhost_8080_ (4)


packages/framework/tests/Feature/PaginatorTest.php
```diff
@@ -0,0 +1,296 @@
...
+        $this->assertSame('page-2.html', $this->makePaginator()->setCurrentPage(1)->next());
+    }
+
+    public function testPreviousAndNextMethodsWithBaseRouteSet()
+    {
+        $pages[1] = new VirtualPage('pages/page-1');
+        $pages[2] = new VirtualPage('pages/page-2');
+        $pages[3] = new VirtualPage('pages/page-3');
+        $pages[4] = new VirtualPage('pages/page-4');
+        $pages[5] = new VirtualPage('pages/page-5');
```
caendesilva, line 201: I wonder if we could add a helper method that generates the paginated virtual pages / blade pages (using a specified template) this could then be used in a provider when booting to inject those pages into the page collection.

caendesilva:
I wonder if we could add a helper method that generates the paginated virtual pages / blade pages (using a specified template) this could then be used in a provider when booting to inject those pages into the page collection. Something for another PR.

PR #833, opened 2023-01-14
Clean up PublicationFieldTypes enum

caendesilva:
**Major changes:**

- Rename PublicationFieldTypes enum case Image to Media
- Sort enum cases by ease of access (more or less)
- Remove numeric rule from integer field type
- Add string to default media type rules


packages/publications/src/PublicationFieldTypes.php
```diff
@@ -58,8 +58,8 @@ public static function getRules(self $type): array
             self::Boolean => [new BooleanRule],
             self::Integer => ['integer', 'numeric'],
```
caendesilva, line 59: I don't think we need numeric here

caendesilva: Removing this makes packages/publications/tests/Feature/MakePublicationCommandTest.php:538 fail.

caendesilva: Confirmed in a real console that strings cannot be passed as integers:
image

False positive?

caendesilva: Ah, the message becomes different. e8b22afbb28a03fa403d20f1dae916544e35331f

PR #855, opened 2023-01-22
Minor code refactors and cleanups

caendesilva:


caendesilva:
Visual change error is false positive caused by unresolvable image

PR #873, opened 2023-01-25
Breaking: Separate internal application files from project structure 

caendesilva:
## About 

This PR is seperating files supporting the underlying Laravel (Zero) application from the actual project structure. This PR **is <ins>breaking</ins>**. The idea behind these changes is to put things that the user usually does not need to edit out of the way so they can focus on what actually matters. Resolves https://github.com/hydephp/develop/issues/874 

### Major Changes
- [x] https://github.com/hydephp/develop/pull/872
- [x] https://github.com/hydephp/develop/pull/890
- [x] https://github.com/hydephp/develop/pull/894
- [x] https://github.com/hydephp/develop/pull/946
- [x] https://github.com/hydephp/develop/pull/947
- [x] https://github.com/hydephp/develop/pull/949
- [x] https://github.com/hydephp/develop/pull/950
- [x] https://github.com/hydephp/develop/pull/978
- [x] https://github.com/hydephp/develop/pull/983
- [x] https://github.com/hydephp/develop/pull/984

#### Additional changes
- [x] Break out application logo from the app config 99f177af2a64cd5a277dbf448d8cd5255742bf8d
- [x] Use a more graceful method for re-enabling package auto-discovery 32b483ab95487e7aa0af92828f85fbd7cc58d350


caendesilva:
See if we can move config/app.php to app/config.php

packages/framework/src/Foundation/ConsoleKernel.php
```diff
@@ -18,6 +18,16 @@ protected function bootstrappers(): array
 
         array_splice($bootstrappers, 5, 0, LoadYamlConfiguration::class);
 
-        return $bootstrappers;
+        // Since we store our application config in `app/config.php`, we need to replace
```
caendesilva, line 21: Since the removed config files like view and cache are now inlined and dependent on a service provider registration we should put that in the bootstrapper.

caendesilva: Resolved in https://github.com/hydephp/develop/pull/978

PR #876, opened 2023-01-27
Add a command to automatically change source directories

caendesilva:
Adds a command that can be used to quickly change source directories. When run, Hyde will update the source directory setting in the config file, then create the directory if it doesn't exist, then move all source directories into it.

### DIFF
--- a/packages/framework/src/Console/Commands/ChangeSourceDirectoryCommand.php
+++ b/packages/framework/src/Console/Commands/ChangeSourceDirectoryCommand.php
@@ -66,8 +66,8 @@ class ChangeSourceDirectoryCommand extends Command
     {
         $this->comment('Updating configuration file');
 
-        // We could also inject the current source root value and check if the setting is even present, but we're keeping it simple for now
-        $search = "'source_root' => '',";
+        $current = config('hyde.source_root', '');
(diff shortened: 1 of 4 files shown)


packages/framework/src/Console/Commands/ChangeSourceDirectoryCommand.php
```diff
@@ -0,0 +1,90 @@
...
+namespace Hyde\Console\Commands;
+
+use Hyde\Console\Concerns\Command;
+use Hyde\Facades\Filesystem;
+use Hyde\Hyde;
+
+/**
+ * @see \Hyde\Framework\Testing\Feature\Commands\ChangeSourceDirectoryCommandTest
+ */
+class ChangeSourceDirectoryCommand extends Command
```
caendesilva, line 14: We might want to add a note about this command in the documentation

caendesilva:
Static analysis issues are acceptable for this change.

packages/framework/src/Console/Commands/ChangeSourceDirectoryCommand.php
caendesilva: Resolved in 1ed1bd023dd390f283cac14d89a393642e6782e9

Issue #867, opened 2023-01-23
Bug in upcoming version: Nested documentation sidebar items are hidden


caendesilva:
To retain the current (release) behaviour, `hyde.subdirectories` must be set to `dropdown` or `flat` when using grouped documentation sidebars. This setting should be automatic for documentation pages to preserve BC and intended behaviour.

caendesilva:
Related bug: `_docs/index.md` is no longer hidden either.

PR #882, opened 2023-01-28
Documentation page navigation bugfixes

caendesilva:
Fixes additional problems found in https://github.com/hydephp/develop/issues/867 regarding functionalities lost during previous refactors. This should fix the last issues and will now match the behaviour of the latest release.

The upcoming documentation site has the index page in the sidebar, this should not be.

image


caendesilva:
Unable to repro this on a local environment. Something is probably wrong with the CI (likely outdated configs)

caendesilva:
This was likely caused by the route key not matching the expected. Maybe the dynamic check needs to use the dynamic option.

PR #895, opened 2023-01-31
Create new logic for converting Markdown to plain text for the search index generation

caendesilva:
## About

Takes an entirely new approach that not only creates a more accurate result, but is much much faster. How much faster? About 30-240 times faster. Yeah, I'm very happy about that.

This makes a discernible difference as the entire `php hyde build:search` command is **over ten times faster**. Search index generation for the official docs now take around 120ms instead of around 1500ms.

This code applies to the search index generation, where the source code information is of course not desirable. 

I created a repository where the new output can be compared to the old: https://github.com/caendesilva/new-hyde-search-string-formatter-comparison/commit/5e94ab56ddf02ce4db0e7a2deb2a62362e12be5f.

## Benchmarks

Benchmark run 10 000 times on the sample in the comment above https://github.com/hydephp/develop/pull/895#issuecomment-1410970066 results in an average of `0.228ms`.

Compare this to the old code which needs a whooping `5.923ms` average for the same dataset. I actually had to lower the iteration count to 1000 to manage it. This means it's an ~2700% speed increase, or almost 30 times faster!

Times for first iteration is even more performant. Around `0.260ms` instead of `60ms` (~24000% increase, AKA 240 times faster!)

Normally I'm not that concerned about speed/performance in the static site generator since it by definition only affects build time, but considering that this feature has been among the slowest parts of the build process by far I think this is a very nice result, especially considering the goal of this PR was just to improve the output.

Benchmarking code seen in 541ce61c2c89e6b957ad3c48678a2834566979df and 3cae16ce9b8b91c1f79359c3bdada20af861abd0


caendesilva:
Copilot generated this fixture, pasting it here to save it. Not sure how much of it is plagiarized. 

            # Markdown
            
            ## Headers
            
            # This is an H1
            ## This is an H2
            ### This is an H3
            #### This is an H4
            ##### This is an H5
            ###### This is an H6
            
            ## Emphasis
            
            *This text will be italic*
            _This will also be italic_
            
            **This text will be bold**
            __This will also be bold__
            
            _You **can** combine them_
            
            ## Lists
            
            Unordered
            
            * Item 1
            * Item 2
              * Item 2a
              * Item 2b
            
            Ordered
            
            1. Item 1
            2. Item 2
            3. Item 3
               * Item 3a
               * Item 3b
            
            ## Images
            
            GitHub Logo
            Format: Alt Text
            
            ## Links
            
            http://github.com - automatic!
            [GitHub](http://github.com)
            
            ## Blockquotes
            
            As Kanye West said:
            
            > We're living the future so
            > the present is our past.
            
            ## Inline code
            
            I think you should use an
            `<addr>` element here instead.
            
            ## Syntax highlighting
            
            ```javascript
            function fancyAlert(arg) {
              if(arg) {
                $.facebox({div:'#foo'})
              }
            }
            ```
            
            ## Task lists
            
            - [x] @mentions, #refs, [links](), **formatting**, and <del>tags</del> supported
            - [x] list syntax required (any unordered or ordered list supported)
            - [x] this is a complete item
            - [ ] this is an incomplete item
            
            ## Tables
            
            | Syntax      | Description |
            | ----------- | ----------- |
            | Header      | Title       |
            | Paragraph   | Text        |
            
            ## SHA references
            
            Any reference to a commit’s SHA-1 hash will be automatically converted into a link to that commit on GitHub.
            
            ## Issue references within a repository
            
            Any number that refers to an Issue or Pull Request will be automatically converted into a link.
            
            ## Username @mentions

            Typing an `@` symbol, followed by a username, will notify that person to come and view the comment. This is called an “@mention”, because you’re mentioning the individual. You can also @mention teams within an organization.
            
            ## Automatic linking for URLs
            
            Any URL (like http://www.github.com/) will be automatically converted into a clickable link.
            
            ## Strikethrough
            
            Any word wrapped with two tildes (like ~~this~~) will appear crossed out.
            
            ## Emoji
            
            GitHub supports emoji!
            To see a list of every image we support, check out the Emoji Cheat Sheet.
            
            ## Task Lists
            
            - [x] This is a complete item
            - [ ] This is an incomplete item
            
            ## Nested Lists
            
            1. First list item
               - First nested list item
                 - Second nested list item
                 -

PR #897, opened 2023-02-01
Refactor documentation search to use relative links

caendesilva:
Refactors to use the same route resolver as other site links. The current page is set to the documentation home page so links are relative to that.


caendesilva:
Gonna close this as I think it adds unnecessary complexity. The original problem I wanted to solve was that search links to the documentation home page link to the index page even when pretty URLs are enabled. But I don't think this bloat and complexity is worth it, and I don't think this method solves the right problem.

PR #920, opened 2023-02-04
Make the media directory configurable

caendesilva:
## Custom media directory 🧪

The media directory houses assets like images and stylesheets. The default directory is `_media`, and upon building the site, 
Hyde will copy all files in this directory to `_site/media` (or whatever your configured output and media directories are).

You can change the path to this directory by setting the `media_directory` setting in `config/site.php`.
Note that this change will affect both the source and output directories. For example, if you set the value to `assets`,
all files from `assets` will be copied to `_site/assets`.

If the setting starts with an underscore, that will be removed from the output directory, so files in `_assets` will be copied to `_site/assets`.

>info Note that you will likely need to manually update `webpack.mix.js` so Laravel Mix can compile the assets correctly.

You will of course also need to copy over any existing files from the old directory to the new one.

### DIFF
--- a/config/site.php
+++ b/config/site.php
@@ -127,4 +127,17 @@ return [
 
     'output_directory' => '_site',
 
+    /*
+    |--------------------------------------------------------------------------
+    | Media Directory
+    |--------------------------------------------------------------------------
+    |
+    | This setting specifies the directory where your media files are stored.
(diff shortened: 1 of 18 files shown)


caendesilva:
Manual testing confirms both compiled site and site viewed with the realtime compiler works with customized source directories.

PR #925, opened 2023-02-06
Refactor LocalFeaturedImage getSource method to no longer return relative hyperlink

caendesilva:
Such a dynamic state is better handled when using the value as it makes it more clear where the "magic" happens.

It also makes the code conform a bit more to the existing code comment.

In some cases this change means that you now must use `Hyde::relativeLink($image->source)` instead of just `$image->source`


caendesilva:
Closing as this actually makes sense, as it allows interoperability with the remote featured image link.

PR #959, opened 2023-02-09
Add internal middlewares to the realtime compiler

caendesilva:
## Abstract

Adds a simple middleware feature to the realtime compiler `HttpKernel`.

The PR also refactors the path rewriting from https://github.com/hydephp/develop/pull/958 to use a middleware. This separates out the concerns from the router.

There is currently no public API added for this feature. As such, the middleware types are not explicitly enforced.

## Usage

Middlewares are simple. They are callables that accept the incoming `Request` instance, and returns a modified one. Each middleware gets executed in order, and override the request instance, before being passed to the `Router`.

### Closure middleware

```php
$this->middleware[] = function (Request $request): Request {
    // Do something with the request here...

    return $request;
};
```

### Invokable class middleware

```php
class ExampleMiddleware extends BaseHttpKernel
{
    public function __invoke(Request $request): Request
    {
        // Do something with the request here...

        return $request;
    }
}

class HttpKernel
    /** @var array<class-string<callable>>|array<callable(Request): Request> */
    protected array $middleware = [
        PathNormalizerMiddleware::class,
    ];
}
```

### Invokable class middleware (anonymous)

```php
$this->middleware[] = new class {
    public function __invoke(Request $request): Request {
        // Do something with the request here...

        return $request;
    }
};
```

### DIFF
--- a/packages/realtime-compiler/src/Http/HttpKernel.php
+++ b/packages/realtime-compiler/src/Http/HttpKernel.php
@@ -14,7 +14,7 @@ use Hyde\RealtimeCompiler\Routing\Router;
  */
 class HttpKernel extends BaseHttpKernel
 {
-    /** @var array<callable(Request): Request> */
+    /** @var class-string[]|array<callable(Request): Request> */
     protected array $middleware = [
         //
     ];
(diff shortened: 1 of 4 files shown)


caendesilva:
Using invokable classes adds a ~0.280ms overhead for the class loader. Which is nothing. Especially since they are much more readable.

PR #963, opened 2023-02-10
Add a reading time helper class

caendesilva:
### Abstract

This was a fun one! Thanks to a great suggestion by @rgasch, here's a class that estimates the time a text will take to read! It's still gonna take a bit of tweaking to get the algorithm and public API right though.

### Meta

To adjust multiplication factor, we use the words per minute constant which is set to 240. This means that a 1000 word input results in a 4 minute and 10 second reading time. For context, wordconter.net shows 3 mins and 38 sec, and Grammarly shows 3 min 59 sec. So we're a bit on the high side, but 240 is a very easy number to work with so I chose that. A higher value leads to a shorter estimated reading time.

### Create a new `ReadingTime` instance

```php
// Via constructor
$time = new ReadingTime('Input text string');

// Via static method
$time = ReadingTime::fromString('Input text string');

// Via static method (from file)
$time = ReadingTime::fromFile('path/to/file.txt');
```

### Get the reading time

```php
// Get the reading time in seconds
$time->getSeconds();

// Get the reading time in minutes (rounded down)
$time->getMinutes();

// Get the remaining seconds after the rounded down minutes
$time->getSecondsOver(); // (Perfect for showing after the getMinutes() value)
```

### Get the reading time in a human readable format

```php
// Get formatted reading time
$time->getFormatted(); // 1min, 30sec

// Get formatted reading time with custom sprintf format
// (first argument is the minutes format, second argument is the seconds format)
$time->getFormatted('%d minutes, %d seconds'); // 1 minutes, 30 seconds
```

### Format reading time using a closure

This is perfect if you want to customize the format further, maybe by only pluralizing values that are not 1?

The closure will receive the minutes and seconds as integers and should return a string.

```php
/**
 * @param  int  $minutes
 * @param  int  $seconds
 * @return string
 */
$closure = function (int $minutes, int $seconds): string {
    return "$minutes minutes, $seconds seconds";
};

$time->formatUsingClosure($closure); // 1 minutes, 30 seconds
```

### Extra

You can also get the word count:

```php
$time->getWordCount();
```


packages/framework/src/Support/ReadingTime.php
```diff
@@ -0,0 +1,107 @@
...
+
+    /** @return array<int, float> The minutes and seconds as floats. */
+    protected function getTokenized(): array
+    {
+        $minutes = $this->getMinutesAsFloat();
+        $fMin = floor($minutes);
+        $fSec = ($minutes - $fMin) * 60;
+
+        return [$fMin, $fSec];
+    }
```
caendesilva, line 106: There is no way it's useful to get a fraction of a second here since this is such a rough estimate. I don't even know if there's a point in having float accessors at all. Maybe internal protected one for the minute one.

caendesilva:
Okay I think the class is pretty fully featured now, so now I just want to simplify the public API.

PR #964, opened 2023-02-10
Merge options from site.php config into hyde.php

caendesilva:
### Abstract
Highly experimental PR to merge the site.php config file with the hyde.php one.

#### Step one - Move config settings

Start by copying all the options in the `config/site.php` config to `config/hyde.php`

#### Step two - Update config calls

Next, if you have written custom code, or created/published any Blade views, that calls `config('site.XXX')` you need to replace that with `config('hyde.XXX')` 

This is fairly easy using search+replace across files in VS Code or another modern code editor or IDE. Try searching for `'site.` and you'll find matches needing replacement. See 73143af96c76fdeb89c3fd74883815c919eca948 for an example for what you want.


config/docs.php
```diff
@@ -21,7 +21,7 @@
     |
     */
 
-    'header_title' => config('site.name', 'HydePHP').' Docs',
+    'header_title' => config('hyde.name', 'HydePHP').' Docs',
```
caendesilva, line 24: Reference environment variable instead of configuration option in config/docs.php
```suggestion
    'header_title' => env('SITE_NAME', 'HydePHP').' Docs',
```

caendesilva: Cherry picked to https://github.com/hydephp/develop/pull/970

PR #988, opened 2023-02-12
Create HydeCoreExtension class to replace page class registration feature

caendesilva:


packages/framework/src/Foundation/HydeKernel.php
```diff
@@ -74,6 +74,8 @@ public function __construct(?string $basePath = null)
         $this->setBasePath($basePath ?? getcwd());
         $this->filesystem = new Filesystem($this);
         $this->hyperlinks = new Hyperlinks($this);
+
+        // Todo: load extensions from config
```
caendesilva, line 77: Actually, a well crafted extension should register itself within a service provider. The only other remaining scenario for installing an extension is if a user wants to make their own, and then they can just use the AppServiceProvider or any other custom provider.

packages/framework/src/Foundation/HydeKernel.php
```diff
@@ -66,14 +66,15 @@ class HydeKernel implements SerializableContract
...
-    protected array $extensions = [];
+    protected array $extensions = [HydeCoreExtension::class];
 
     public function __construct(?string $basePath = null)
     {
         $this->setBasePath($basePath ?? getcwd());
         $this->filesystem = new Filesystem($this);
         $this->hyperlinks = new Hyperlinks($this);
+
+        // Todo: load extensions from config
```
caendesilva, line 77:
```suggestion
```

Issue #1009, opened 2023-02-13
Create LocationService helper class


caendesilva:
> Could be improved by having a location service that could also find kebab-case class names in config assuming they resolve to a class in the main pages namespace

```suggestion
            HtmlPage::class => LocationService::getClassInfoFromConfig(class: HtmlPage::class, option: 'source_directories', default_ '_pages'),
            BladePage::class => LocationService::getClassInfoFromConfig(class: BladePage::class, option: 'source_directories', default_ '_pages'),
            MarkdownPage::class => LocationService::getClassInfoFromConfig(class: MarkdownPage::class, option: 'source_directories', default_ '_pages'),
            MarkdownPost::class => LocationService::getClassInfoFromConfig(class: MarkdownPost::class, option: 'source_directories', default_ '_posts'),
            DocumentationPage::class => LocationService::getClassInfoFromConfig(class: DocumentationPage::class, option: 'source_directories', default_ '_docs'),
```

_Originally posted by @caendesilva in https://github.com/hydephp/develop/pull/1007#discussion_r1104211366_

caendesilva:
Closing as helpers can be added into RegistersFileLocations trait for now, and extracted to class if needed later on.

PR #1017, opened 2023-02-13
Update internal page class name resolver for config helper method to support kebab-case class names

caendesilva:
Resolves semantics in https://github.com/hydephp/develop/issues/1009

Could be improved by checking if class exists in our namespace, but don't think the complexity is worth it at this point.


caendesilva:
Realized we could replace the HSP `registerSourceDirectories` call with a loop that iterates through source directory options which could automatically support package directories but I'm not sure it actually adds much value, and I think the verbosity of the current logic is helpful as it makes it the data origin flow more intuitive.

PR #1019, opened 2023-02-13
Refactor InMemoryPages and DynamicPage interface

caendesilva:


packages/framework/src/Pages/InMemoryPage.php
```diff
@@ -27,8 +27,11 @@
  */
 class InMemoryPage extends HydePage implements DynamicPage
 {
+    /** @todo See if we can remove these as they cannot be used */
     public static string $sourceDirectory = '';
+    /** @todo See if we can remove these as they cannot be used */
     public static string $outputDirectory = '';
+    /** @todo See if we can remove these as they cannot be used */
     public static string $fileExtension = '';
```
caendesilva, line 35: Removing the default empty values will require us to update several accessors, some of which are final. But since empty sourceDirectories are not discoverable by 4dce8965c852418a78f97d12cc6430e56b488bb1, this class will still be considered dynamic.

```suggestion
    public static string $sourceDirectory = '';
    public static string $outputDirectory = '';
    public static string $fileExtension = '';
```

PR #1063, opened 2023-02-17
Refactors the HydePage discovery data internals

caendesilva:
### Abstract

When the HydePage class was first designed, all its pages were discoverable. That is no longer the case, so this PR is experimenting with modernizing the base HydePage class by creating an interface that can be used with intersecting type hints against the interface and base class for the places that interact with discoverable pages.

### Major changes

- [x] https://github.com/hydephp/develop/pull/1061
- [x] https://github.com/hydephp/develop/pull/1062
- [x] https://github.com/hydephp/develop/pull/1065
- [x] https://github.com/hydephp/develop/pull/1066
- [x] https://github.com/hydephp/develop/pull/1067
- [x] https://github.com/hydephp/develop/pull/1068
- [x] https://github.com/hydephp/develop/pull/1069
- [x] https://github.com/hydephp/develop/pull/1070
- [x] https://github.com/hydephp/develop/pull/1071
- [x] `HydePage::sourceDirectory()` now falls back to the source root when property is not initialized
- [x] `HydePage::outputDirectory()` now falls back to the empty string when property is not initialized
- [x] `HydePage::fileExtension()` now falls back to the empty string when property is not initialized


caendesilva:
Output directories must of course be in all page types. I'm stupid. This Caen, this is why you don't try to code when you have a fever and should be resting.

caendesilva:
Okay I figured out a much smarter (or way less dumb) way to do this. The problem I was trying to solve was that some pages aren't discoverable and shouldn't have source paths, but pragmatically (and almost semantically) it doesn't matter if they just return empty strings or the site roots.

caendesilva:
RE public/protected+accessor properties

I realized that I probably don't need to spend to much time thinking about this as only very very few people will ever modify these properties directly. The few people that want to customize source directories can just use the config which is the "official"/recommended way of doing it. If someone for some reason wants to do the change via the code directly in a service provider or similar they (I assume/hope) are knowledgeable enough to debug things when they go wrong. So I am probably overthinking this part. As long as we make sure the official way is properly done, any unsupported hacks should be on the responsibility of that developer, not us.

PR #1067, opened 2023-02-17
Breaking: Break out discoverable properties from HydePage base class

caendesilva:
Changes the HydePage classes to use the Discoverable trait added in https://github.com/hydephp/develop/pull/1066 so that only discoverable pages use the trait, not the base HydePage class.


caendesilva:
I swear PHP is trolling me sometimes. Apparently we can't redefine trait properties, so new base class it is! At this point I'm starting to question if this whole line of changes is worth it, but I'm learning stuff so I'll proceed, but may not merge into master later on.

caendesilva:
Handling merge conflicts in target branch

PR #1068, opened 2023-02-17
Convert trait into new base class for discoverable pages

caendesilva:
I swear PHP is trolling me sometimes. Apparently we can't redefine trait properties, so new base class it is! At this point I'm starting to question if this whole line of changes is worth it, but I'm learning stuff so I'll proceed, but may not merge into master later on.

This change is not breaking as the trait had not yet been merged into master. This does however partially revert https://github.com/hydephp/develop/pull/1066 as the code there is merged into the new base class.


caendesilva:
Handling merge conflicts in target branch

PR #1079, opened 2023-02-20
Fix bug inconsistent front matter serialization for empty publication field input

caendesilva:
This means that skipped fields will still be included in the saved output even if they're empty.

Fixes https://github.com/hydephp/publications/issues/6


caendesilva:
Closing as stale

PR #1093, opened 2023-02-21
Aggressively simplify the featured image classes

caendesilva:
This provides a nice feature, but there's honestly not enough reason to use three classes for such a scoped feature. Hence this (obviously breaking) refactor to aggressively simplify these classes, merging in the two child classes into the base abstract class.


caendesilva:
Thinking it might simplify the API if we merge `image.path` and `image.url` into a single `image.source` that determines local/remote state depending if it starts with `http` or `//`. Would be rather breaking though but may be worth it.

PR #1094, opened 2023-02-22
Breaking: Merge blog post featured image source front matter properties

caendesilva:
## Abstract

Implements the proposed change in https://github.com/hydephp/develop/pull/1093#issuecomment-1439109372 to simplify the API by merging `image.path` and `image.url` into a single `image.source` (or just the `image` one when not using array) property that determines local/remote state depending on whether it starts with `http` or `//`. 

Although this is a breaking change, I believe it is worth it. This change also aligns with the internal data structure, where both input options are stored in the `$source` property. Furthermore, we no longer need to specify that one property overrides the other.

In other words, both `image.url` and `image.path` are now `image.source`, which equals the behaviour of when using `image: somestring`

### Flat syntax: Before & After

No change here.

```yaml
image: image.jpg
```
```yaml
image: https://cdn.example.com/image.jpg
```

### Array syntax: Before

```yaml
image:
  path: image.jpg
```
```yaml
image: 
  url: https://cdn.example.com/image.jpg
```

### Array syntax: After

```yaml
image:
  source: image.jpg
```
```yaml
image: 
  source: https://cdn.example.com/image.jpg
```

### DIFF
--- a/docs/creating-content/blog-posts.md
+++ b/docs/creating-content/blog-posts.md
@@ -186,11 +186,11 @@ will override all the values in the `authors` config entry.
 ### Image
 
 ```yaml
-image: image.jpg # Expanded by Hyde to `_media/image.jpg` and is resolved automatically
-image: https://cdn.example.com/image.jpg # Full URL starting with `http(s)://`)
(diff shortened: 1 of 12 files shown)


caendesilva:
The hydephp.com blog post for this needs to be updated

PR #1099, opened 2023-02-22
First (very raw) version of the logic to generate tag files.

rgasch:
I am almost embarrassed to call this a pull request, but this is the core logic required to generate tag files (1 summary/index page + a list page for every tag). You'll have to do quite a bit of refactoring to properly fit this into the hyde codebase, but the logic seems to be correct).


caendesilva:
Great! Thanks, I'll start working against this branch tomorrow. Don't worry about the test static analysis test failing, I apparently need to update it so it handles forks gracefully.

caendesilva (approved):
Looks good to me now!

caendesilva:
Merged into https://github.com/hydephp/develop/pull/685. Thanks!

PR #1103, opened 2023-02-23
Add top level meta field to publication

rgasch:


packages/publications/tests/Feature/PublicationTypeTest.php
```diff
@@ -423,7 +424,8 @@ public function testJsonRepresentationWithDefaultValues()
                 "sortField": "__createdAt",
                 "sortAscending": true,
                 "pageSize": 0,
-                "fields": []
+                "fields": [],
+                "meta": []
```
caendesilva, line 428: @rgasch what do you think about only storing the meta array when it's not empty? That means it wouldn't be present in defaultly generated files, but could easily be added by the user? 

Pros/cons IMO
+ Cleaner schema
- Extra steps for users

Of course the field will still be documented either case.

packages/publications/src/Models/PublicationType.php
```diff
@@ -57,6 +57,9 @@ class PublicationType implements SerializableContract
     /** The number of publications to show per paginated page. Set to 0 to disable pagination. */
     public int $pageSize = 0;
 
+    /** Generic array field which can be used to store additional data as needed. */
+    public array $meta = [];
```
caendesilva, line 61: What do you think about renaming this (and accessors) to `$metadata`?

```suggestion
    public array $metadata = [];
```

rgasch: My personal view is that since every class variable holds data, meta"DATA" is sort of superfluous. But if you want me to make this change, let me know and I'll do it.

caendesilva: I see your point, for me it feels that "meta" is an abbreviation of "metadata" in this case. So as you are not opposed to the idea I think we should use "metadata". It looks a bit "cut off" IMO.

caendesilva:
Merging into https://github.com/hydephp/develop/pull/685. Thanks!

PR #1104, opened 2023-02-23
Return matter as collection

rgasch:
Convert FrontMatter data to use Collection. This allows you to access FM data items like this:

$publication->matter->{$fieldname}

### DIFF
--- a/packages/framework/src/Markdown/Models/FrontMatter.php
+++ b/packages/framework/src/Markdown/Models/FrontMatter.php
@@ -8,6 +8,7 @@ use Hyde\Framework\Actions\ConvertsArrayToFrontMatter;
 use Hyde\Support\Concerns\Serializable;
 use Hyde\Support\Contracts\SerializableContract;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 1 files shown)


caendesilva:
Can you share some code examples of how this change improves the DX in comparison to the current structure (and if we added a ->collect() method chain)?

caendesilva:
@rgasch I'm already able to use `$page->matter->{$property}` as the `FrontMatter` class implements the magic `__get` method. 

All HydePage objects have both a class property called `$matter` (which is an instance of the `FrontMatter` class), and a `->matter()` method which calls the `->get()` method on said instance. (The latter feature is implemented in `InteractsWithFrontMatter.php`)

Upon taking a look at the code again, I see that the `$data` property in the FrontMatter class is public. I don't think it needs to be since it has accessors. @rgasch what do you think about protecting it?

caendesilva:
Closing as `$publication->matter->{$fieldname}` already works from my testing as the FrontMatter object implements the magic __get method.

PR #1113, opened 2023-02-24
Add final InMemoryPage source path accessor override 

caendesilva:
InMemoryPages can, by definition, not have a source path. This PR adds a final method override to return an empty string `''` to force a consistent behaviour.


caendesilva:
Did not work as I expected. Closing.

PR #1120, opened 2023-02-26
Adding breadcrumb component.

rgasch:


caendesilva (commented):
Thanks of the PR! Code looks good, we'll see if we actually need a component class if we don't have logic in it. Do you want to add some tests for this? Otherwise I'll do it a bit later today.

caendesilva:
Okay I can't properly change the target branch without risking messing up yours with a force push so I'll merge this into https://github.com/hydephp/develop/pull/685 and then cherry pick it from there to master.

PR #1138, opened 2023-02-27
Refactor publications into the kernel container

caendesilva:
Fixes https://github.com/hydephp/publications/issues/9


caendesilva:
Just a fun sidenote, rebooting the kernel during testing the so far 40 times takes 244ms

PR #1140, opened 2023-02-28
Adding relatedPublications component.

rgasch:


packages/publications/src/Models/PublicationPage.php
```diff
@@ -23,6 +23,7 @@ class PublicationPage extends Concerns\BaseMarkdownPage
 {
     use ValidatesExistence;
 
+    // Fixme: can we make this private or protected without breaking other stuff?
```
caendesilva, line 26: There are actually a lot of public properties that can be protected. If you have time, I'd love to get your design thoughts on this issue if you take a look at the code in the core framework package. Mainly thinking about the page classes like `HydePage` and so

packages/publications/src/Models/PublicationPage.php
```diff
@@ -46,6 +47,11 @@ public function compile(): string
         return $this->renderComponent();
     }
 
+    public function getType(): PublicationType
+    {
+        return $this->type;
+    }
```
caendesilva, line 53: I'll cherry pick this directly to the publications branch as that can be reviewed independently of this PR; unless you want credit through a dedicated PR.

packages/framework/src/Framework/Views/Components/RelatedPublicationsComponent.php
```diff
@@ -0,0 +1,104 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Hyde\Framework\Views\Components;
```
caendesilva, line 5: This class should be in the Publications package so the core package functions without it, but I can fix that if you want.

packages/framework/src/Framework/Views/Components/RelatedPublicationsComponent.php
```diff
@@ -0,0 +1,104 @@
...
+
+use Hyde\Hyde;
+use Hyde\Publications\Models\PublicationPage;
+use Hyde\Publications\PublicationService;
+use Illuminate\Contracts\View\Factory;
+use Illuminate\Contracts\View\View;
+use Illuminate\Support\Collection;
+use Illuminate\View\Component;
+
+class RelatedPublicationsComponent extends Component
```
caendesilva, line 15: This class is adding a considerable amount of complexity. Can you try adding some tests for it?

caendesilva:
@rgasch Just wanted to let you know that in case this is not merged into the actual package, it's always easy to add custom components to individual Hyde projects (the same way you do in Laravel)

packages/publications/src/Models/PublicationPage.php
rgasch: Do what's easiest ... just cherrypick it ..

packages/framework/src/Framework/Views/Components/RelatedPublicationsComponent.php
rgasch: Please fix that. Thank you.

packages/framework/src/Framework/Views/Components/RelatedPublicationsComponent.php
rgasch: Yes. I think the way to go is to refactor it into a few dedicated sub-methods which can then be tested for correctness per method and then test the entire thing with some fake data. I'll take a look at adding tests later this week.

rgasch:
> @rgasch Just wanted to let you know that in case this is not merged into the actual package, it's always easy to add custom components to individual Hyde projects (the same way you do in Laravel)

I think that this is a very useful addition that can be used by anybody using the tags feature ... as such, if possible, lets try to get it merged ...

packages/framework/src/Framework/Views/Components/RelatedPublicationsComponent.php
caendesilva: I think that sounds great.

packages/publications/src/Models/PublicationPage.php
caendesilva: Will do.

caendesilva:
> > @rgasch Just wanted to let you know that in case this is not merged into the actual package, it's always easy to add custom components to individual Hyde projects (the same way you do in Laravel)
> 
> I think that this is a very useful addition that can be used by anybody using the tags feature ... as such, if possible, lets try to get it merged ...

Deal. If this was for the core package I would be more hesitant to add something so specific, but since this is an extension I think we have more free room to roam around.

packages/publications/src/Models/PublicationPage.php
caendesilva: Resolved in https://github.com/hydephp/develop/pull/1148

caendesilva:
@rgasch Heads up, your branch was pretty behind, so I merged the latest changes. I'm gonna move the namespaces (as mentioned in https://github.com/hydephp/develop/pull/1140/commits/e94cf5b3246d8585e12af187ccd06b33a0333691#r1119891257) so you'll have to re-pull those changes in a bit as well.

caendesilva:
Ignore the test fails, I've messed up the workflow. Will fix in a bit if I have time today

caendesilva:
I will be renaming the component to be kebab-case to match the other views, I will also be moving the files. Just heads up that you need to pull the files when I'm done.

caendesilva:
Merged into https://github.com/hydephp/develop/pull/685. Awaiting further refactors there.

PR #1152, opened 2023-03-01
Finalize the BuildTasks API

### LINKED ISSUE #1150: Support overloading build tasks
> On our end, we should add a feature to replace build tasks (if a user build task has the same class base name as a built in one, we only call the user's)

_Originally posted by @caendesilva in https://github.com/hydephp/publications/issues/10#issuecomment-1449612990_

### LINKED ISSUE #1151: BuildTasks constructors need to be consistent
Since there is no way to inject additional dependencies into the build task constructors (nor is there a need for it IMO) the constructors should either have a consistent constructor tag, or make the constructor final.

### LINKED ISSUE #1153: Support PreBuildTasks
The build tasks are kindof in a limbo, where they are named BuildTasks, but we also use the name PostBuildTasks. To be honest, we should probably support PreBuildTasks as well.

### LINKED ISSUE #1154: Refactor BuildTaskService to collect build task classes before running them
This will make it more transparent which tasks are going to run. We could also key them in an internal array to provide support for https://github.com/hydephp/develop/issues/1150

### LINKED ISSUE #1155: Convert BuildTaskService to singleton
It may be beneficial to refactor the BuildTaskService to be a singleton which would help with https://github.com/hydephp/develop/issues/1154.

We'd probably want it in the Laravel service container. We should decouple the output constructor to be a setter so in the build command we can run the following:

```php
$service = app(BuildTaskService::class);
$service->setOutput($output);
```

### LINKED ISSUE #1156: Decouple BuildTaskService output assignment from constructor
We should decouple the output constructor to be a setter so in the build command we can run the following:

```php
$service = app(BuildTaskService::class);
$service->setOutput($output);
```

_Originally posted by @caendesilva in https://github.com/hydephp/develop/issues/1155_

### LINKED ISSUE #1158: Invert the run/handle methods of build tasks
Since build tasks are in a sense "micro commands" it is more semantic to have the handle method be the abstract method used to provide the task logic, and the run method be the one called by the service.

This would require an inversion of how the exit code is handled too, but I think it would be more intuitive as it would be changed so the child method returns a value instead of needing a class property for the exit code.

### LINKED ISSUE #1160: Refactor output directory cleaning to be a PreBuildTask
Now that Hyde will finally support PreBuildTasks, the following BuildService method is a perfect candidate for refactoring:

```php
public function cleanOutputDirectory(): void
{
    if (config('hyde.empty_output_directory', true)) {
        $this->warn('Removing all files from build directory.');

        if ($this->isItSafeToCleanOutputDirectory()) {
            array_map('unlink', glob(Hyde::sitePath('*.{html,json}'), GLOB_BRACE));
            File::cleanDirectory(Hyde::siteMediaPath());
        }
    }
}
```

### LINKED ISSUE #1163: Update the BuildTasks API documentation
Entire article needs to be rewritten to fit the breaking changes in https://github.com/hydephp/develop/pull/1152

### LINKED ISSUE #1174: Move framework build tasks out of feature namespace
These actions are not related to the public feature API

### LINKED ISSUE #1175: Site build command should not be responsible for registering build services
Either do this in a provider like a build service provider, or in the build task service itself.

### LINKED ISSUE #1176: Add a BuildServiceProvider
> Actually, since performance of 20ms does not matter at all at this point, I'm going to go with what's best architecture, and I think it's better to handle registrations in a provider as that's what they are for. Question is, do we do a BuildTaskServiceProvider or just a BuildServiceProvider in general? Think the latter.

_Originally posted by @caendesilva in https://github.com/hydephp/develop/issues/1175#issuecomment-1452115815_

### LINKED ISSUE #1177: Remove exit code feature from build tasks since they aren't used anywhere
We never use the exit codes, and I'm not sure where and how we'd want to do that. It was added as a way to handle exceptions, I think it's better to add an option that controls whether to throw or catch.

Otherwise we'd have to pass back the exit code reference to the command but I don't think that's worth the complexity.

caendesilva:
### Abstract

Revamps the entire BuildTasks API to ensure it's polished for the 1.0 release. This PR contains several breaking changes. The changes adds quite a bit of abstraction and verbosity, but I do think the API is very nice and fluent to work with.

Note that a large part of the file diff (lines added) is because of the an extensive unit test suite has been added. Another drawback of this is that the build command is a bit less clear, but since the BuildTask API is much more likely to be used by end users compared to the command, I think that is okay.

### Resolves the following

- [x] Fixes #1163
- [x] Fixes #1153
- [x] Fixes #1150
- [x] Fixes #1151
- [x] Fixes #1154
- [x] Fixes #1156
- [x] Fixes #1158
- [x] Fixes #1155
- [x] Fixes #1160
- [x] Fixes #1174
- [x] Fixes #1175
- [x] Fixes #1177
- [x] Fixes #1176


packages/framework/src/Console/Commands/BuildSiteCommand.php
```diff
@@ -77,7 +78,13 @@ protected function runPreBuildActions(): void
...
     public function runPostBuildActions(): void
     {
-        $service = new BuildTaskService($this->output);
+        $service = app(BuildTaskService::class);
+        $service->setOutput($this->output);
+
+        $service->registerIf(GenerateSitemap::class, $this->canGenerateSitemap());
+        $service->registerIf(GenerateRssFeed::class, $this->canGenerateFeed());
+        $service->registerIf(GenerateSearch::class, $this->canGenerateSearch());
+        $service->registerIf(GenerateBuildManifest::class, config('hyde.generate_build_manifest', true));
```
caendesilva, line 87: We could also make so that these are registered in the config, as an alternate way of overloading (see https://github.com/hydephp/develop/issues/1150) but I think the use case for this is relatively slim, and the automatic overloading being implemented is nicer.

packages/framework/src/Framework/Services/BuildTaskService.php
```diff
@@ -4,89 +4,114 @@
...
 {
-    /**
-     * Information for package developers: This offers a hook for packages to add custom build tasks.
-     * Make sure to add the fully qualified class name to the array and doing so by merging the array, not overwriting it.
-     *
-     * @var array<class-string<\Hyde\Framework\Features\BuildTasks\BuildTask>>
-     */
-    public static array $postBuildTasks = [];
+    /** @var array<string, class-string<\Hyde\Framework\Features\BuildTasks\BuildTask>> */
+    protected array $buildTasks = [];
```
caendesilva, line 32: Refactor to store instantiated objects instead of class-strings

This makes it so much easier to mock them, it also provides native type support

packages/framework/src/Framework/Services/BuildTaskService.php
```diff
@@ -4,89 +4,116 @@
...
-    public function run(string $task): static
+    /** @param  class-string<\Hyde\Framework\Features\BuildTasks\BuildTask>  $class */
+    public function registerTask(string $class): void
     {
-        $this->runTask(new $task($this->output));
+        $task = new $class($this->output);
+
+        if (! $task instanceof BuildTask) {
+            throw new InvalidArgumentException("BuildTask [$class] must extend the HydeBuildTask class.");
+        }
```
caendesilva, line 77: Throw an exception if the task does not implement a runner interface as the task will never run otherwise, which could be confusing for the developer. Throwing makes it clear that something is indeed wrong.

caendesilva: We could also add a warning if both are implemented as the same handle method is run regardless of this, but I don't think that matters enough. We'll put something in the docs.

caendesilva: Should we default to implements RunsAfterBuild?

packages/framework/src/Framework/Services/BuildTaskService.php
```diff
@@ -4,89 +4,122 @@
...
+            }
+        }
     }
 
-    public function run(string $task): static
+    /** @param  \Hyde\Framework\Features\BuildTasks\BuildTask|class-string<\Hyde\Framework\Features\BuildTasks\BuildTask>  $task */
+    public function registerTask(BuildTask|string $task): void
     {
-        $this->runTask(new $task($this->output));
+        $task = is_string($task) ? new $task($this->output) : $task;
```
caendesilva, line 73: We can't set output here since output might not be set yet. It needs to be handled at task run time

caendesilva: Additionally, with 33235ab006cb899a4316a0babcb285be739b86c9 this means that already constructed objects can't get the output-

caendesilva: Resolved in ad288caea16f2c7137ed153a49e8356e3cf7498d

caendesilva:
Okay now that it's another day and I'm taking a look at this with fresh eyes, I'm going to convert the interfaces to base classes. I think that is clearer and requires less explanation.

Issue #1164, opened 2023-03-01
Documentation pages cannot have flattened output paths


caendesilva:
This felt like a good idea at the time, but considering that only one of the following pages are rendered, this clearly doesn't work:

```
_docs/group1/foo.md
_docs/group2/foo.md
```

caendesilva:
Somewhat related: Documentation pages with the same navigation labels (but differing categories) are not shown

caendesilva:
Reimplementing this feature as configurable in https://github.com/hydephp/develop/pull/1171

This means that the breaking change in https://github.com/hydephp/develop/pull/1166 is reverted. To solve the original issue, the fix is either to change the new config setting, or to change the filenames. For example:

```
_docs/group1/group1.md
_docs/group2/group1.md
```

And then just use custom labels.

--- 

However, this would be nice:

```
_docs/group1/index.md
_docs/group2/index.md
```

Maybe we can make so nested index pages are not flattened?

PR #1166, opened 2023-03-01
Breaking: Remove flattened output documentation paths

caendesilva:
Removes the UsesFlattenedOutputPaths trait, and the overall feature, as it breaks pages with the same name in different directories. Fixes https://github.com/hydephp/develop/issues/1164

This is unfortunately likely lead to URL breaks in built sites.


caendesilva:
This broke search. Search index link must be updated to resolve the file based on the root docs path-

caendesilva:
Okay gonna try to see if I can add back this as a simple configurable feature.

PR #1185, opened 2023-03-02
Refactor RelatedPublicationsComponent

rgasch:
This is the promised refactoring of the RelatedPublicationsComponent; it should be more readable and easily understandable than the original version. j

Sorry for the superfluous registration of the component in the wrong file. It tried to reset/undo the commit for this (which seems to have worked locally) but github refuses to honor it. So please undo this 1 line change after merging this PR.

### DIFF
--- a/packages/framework/src/Publications/Views/Components/RelatedPublicationsComponent.php
+++ b/packages/framework/src/Publications/Views/Components/RelatedPublicationsComponent.php
@@ -32,17 +32,17 @@ class RelatedPublicationsComponent extends Component
     protected function makeRelatedPublications(int $max = 5): Collection
     {
         // Get current publicationType
-        $currHydePage = Hyde::currentRoute()->getPage();
-        $publicationType = $currHydePage->getType();
-        if (! $publicationType) {
+        $currentHydePage = Hyde::currentRoute()->getPage();
+        $publicationType = $currentHydePage->getType();
+        if (!$publicationType) {
             return collect();
         }
 
         // Get the tag fields for the current publicationType -> exit if there aren't any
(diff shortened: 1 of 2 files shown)


caendesilva:
I think the reason you are having issues is that you are pushing from the same branch. GitHub only accepts one pull request per source branch as far as I can tell. So if you make more than one PR, you need to do it from a different branch on your end.

Please see if you can push a revert commit for [0f05449](https://github.com/hydephp/develop/pull/1185/commits/0f054497fd896fbe8c6527f0c4dfc172ff02d241).

packages/framework/src/Publications/Views/Components/RelatedPublicationsComponent.php
```diff
@@ -25,72 +29,101 @@ public function render(): Factory|View
         return view('hyde-publications::components.related-publications');
     }
 
-    protected function makeRelatedPublications(int $max = 5): array
+    protected function makeRelatedPublications(int $max = 5): Collection
     {
         // Get current publicationType
-        $currPage = Hyde::currentRoute()->getPage();
-        $publicationType = $currPage->getType();
+        $currHydePage = Hyde::currentRoute()->getPage();
```
caendesilva, line 35: Please see if you can expand all the abbreviated variable names.

caendesilva (changes requested):

packages/framework/src/Publications/Views/Components/RelatedPublicationsComponent.php
rgasch: Done!

caendesilva (approved):
Still needs some tests, but that can be in a different PR.

PR #1193, opened 2023-03-03
Finalize foundation collections

caendesilva:
### Abstract

Polishes and finalizes the foundation collections in preparation for the 1.0 release.

### Issues

- [x] Fixes https://github.com/hydephp/develop/issues/1199
- [x] Fixes https://github.com/hydephp/develop/issues/1194
- [x] Fixes https://github.com/hydephp/develop/issues/1200
- [x] Fixes https://github.com/hydephp/develop/issues/1196

### Breaking/significant changes

- Moves collection accessors to the facade classes to limit scope of the classes to focus on data discovery
- Make the BaseFoundationCollection::boot method final
- Change BaseFoundationCollection::runDiscovery to return void instead of self
- Change BaseFoundationCollection::runExtensionCallbacks to return void instead of self
- Make FileCollection::addFile method return void
- Make PageCollection::addPage method return void
- Make RouteCollection::addRoute method return void

### DIFF
--- a/packages/framework/src/Foundation/Concerns/BaseFoundationCollection.php
+++ b/packages/framework/src/Foundation/Concerns/BaseFoundationCollection.php
@@ -49,6 +49,7 @@ abstract class BaseFoundationCollection extends Collection
     {
         try {
             $this->runDiscovery();
+            $this->runExtensionCallbacks();
         } catch (Throwable $exception) {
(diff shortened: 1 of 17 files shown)


caendesilva:
With this, the Route facade and the Routes facade overlap. I think they can be merged.

Issue #1204, opened 2023-03-04
Break out MediaFile discovery from the Kernel/FoundationCollection


caendesilva:
In all honesty, MediaFile objects do not need to be auto-discovered at all, considering they are only ever used in the build loop. Since they are never further processed, that complexity is not warranted.

PR #1205, opened 2023-03-04
Aggressively simplify MediaFile/FileCollection discovery

caendesilva:
In all honesty, MediaFile objects do not need to be auto-discovered at all, considering they are only ever used in the build loop. Since they are never further processed, that complexity is not warranted. Nor do they have any need to be cached. Fixes https://github.com/hydephp/develop/issues/1204

### DIFF
--- a/packages/framework/src/Foundation/Facades/Files.php
+++ b/packages/framework/src/Foundation/Facades/Files.php
@@ -46,12 +46,6 @@ class Files extends Facade
         return static::getFacadeRoot()->where(fn (ProjectFile $file): bool => $file instanceof SourceFile);
     }
 
-    /** @deprecated */
-    public static function getMediaFiles(): FileCollection
-    {
-        return MediaFile::all();
-    }
-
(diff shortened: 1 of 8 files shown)


caendesilva:
For context, here are the some benchmarks on the time it takes to run the new helper once, each value is the measured time in milliseconds belonging to a test run with varying amounts of files.

1.4037, 3.0746, 1.535, 1.6649, 1.3965, 1.4494, 1.5352, 1.4993, 1.5649, 1.6762, 1.6937, 1.7277, 1.7275, 1.4097, 0.3393, 2.0069, 2.1428, 0.216, 0.6956, 0.2081, 0.6212, 1.5015, 1.4237, 1.9058, 2.1781, 1.4807, 1.5292, 1.3859, 1.3822, 1.3999, 1.403, 1.4447, 1.3839, 2.0084, 1.4949, 1.3769, 1.4086, 1.3723, 1.4625, 1.41, 1.3693, 1.3818, 1.4202, 1.4831

Issue #1198, opened 2023-03-04
Merge Route facade into Foundation Routes facade


caendesilva:
With this, the Route facade and the Routes facade overlap. I think they can be merged.

_Originally posted by @caendesilva in https://github.com/hydephp/develop/issues/1193#issuecomment-1454690421_

caendesilva:
Whoops didn't mean to close

caendesilva:
Actually, if I had to choose one over the other, I think the Route facade is the one to keep. The only problem this would be solving (as merging adds complexity to a scoped class) is to IDE autocompletion conflicts between the Route model and the Route facade, and I don't think that's that huge of a problem. However, it can be simplified a little bit.

caendesilva:
Actually tbh Routes::get() makes 95% as much semantic sense as Route::get(). Maybe that's enough to justify a merge.

PR #1215, opened 2023-03-06
Merge Route facade into Foundation Routes facade

caendesilva:
Pull Request Description:

This pull request is aimed at addressing the issue raised in #1198 regarding the overlap between the Route facade and the Routes facade. After reviewing the issue and its comments, we have decided to merge both facades to reduce complexity.

The current overlap between the facades has caused confusion and created conflicts in IDE autocompletion, which can be solved by merging the two facades. While there were concerns raised about the namespace differences between the facades, we believe that the semantic sense of using Routes::get() is close enough to Route::get() to justify the merge.

We have made the necessary changes to merge the Route facade into the Foundation Routes facade, simplifying the codebase and avoiding unnecessary duplication. The merge also makes it easier for developers to work with the facade in the future.

Overall, we believe that merging the facades is a positive change that will improve the codebase's maintainability and reduce potential confusion for developers. We look forward to feedback and suggestions on this proposed change.


packages/framework/resources/views/components/navigation/navigation-brand.blade.php
```diff
@@ -1,3 +1,3 @@
-<a href="{{ Route::get('index') }}" class="font-bold px-4" aria-label="Home page">
+<a href="{{ \Hyde\Foundation\Facades\Routes::get('index') }}" class="font-bold px-4" aria-label="Home page">
```
caendesilva, line 1:
```suggestion
<a href="{{ Routes::get('index') }}" class="font-bold px-4" aria-label="Home page">
```

PR #1218, opened 2023-03-06
Finalize page model classes and related helpers

caendesilva:
Major cleanups (including breaking changes) relating to the page model classes

As this is part of the sprint to prepare HydePHP 1.0 there are breaking changes here, and they are not all going to be documented as the BC promise has not yet kicked in.


caendesilva (commented):
Reviewed, looks good to me.

Issue #1213, opened 2023-03-05
Normalize static constructor naming


caendesilva:
## Abstract - Make vs Create

We use both `make` and `create` for static constructors. I thought about having one of them be for pure methods performing the same semantics as the standard constructors, but that just adds complexity and needs explaining that is better resolved using `#[Pure]` annotations.

## Search Comparison

Inspired by the Wikipedia notability checks, I went to the always trustworthy Google test to see what had the most results. Here's the breakdown.

| Search Query                                 | Results          |
| -------------------------------------------- | -----------------|
| "public static function make"                | About 5,450      |
| "public static function make" + "php"        | About 5,310      |
| "public static function make" + "laravel"    | About 2,730      |
| "public static function create"              | About 62,400     |
| "public static function create" + "php"      | About 55,800     |
| "public static function create" + "laravel"  | About 6,990      |

## Conclusion

In conclusion, while both `make` and `create` can be used for static constructors, the usage of `create` appears to be much more prevalent in the programming world, particularly in PHP and Laravel. Therefore, it may be wise to stick with the convention of using `create` for static constructors, but also keep in mind the possibility of using `#[Pure]` annotations for pure methods to avoid unnecessary complexity. Ultimately, and speaking generally, the choice between `make` and `create` may come down to personal preference and the specific context in which they are being used.

Considering my Twitter network has a rather strong preference for `make`, that weights a lot in the context of HydePHP. I think I will try using `make` for pure constructors, and `create` for ones being persisted or have strong side effects.

## Examples

If you're not sure what I mean by static constructors, here's an example using both wordings:

```php

class Author
{
    public function __construct(string $username)
    {
        $this->username = $username;
    }

    public static function create(string $username): static
    {
        return new static($username);
    }
    
    public static function make(string $username): static
    {
        return new static($username);
    }
}
```

PR #1220, opened 2023-03-07
Refactor codebase to use consistent naming formats and scoped consistent public API

caendesilva:
## Abstract

Fixes https://github.com/hydephp/develop/issues/1213 and cleans up and finalizes surrounding code by normalizing code names, and scoping down and normalizing the public API, protecting code that does not need to be public, so that the entry points are simpler to manage. Also marks properties that do not support being persisted as readonly.

### Naming Conventions for Action Classes

Actions: `execute()`
Actions with static entry point: `handle()`
Action classes that are called as functions: `__invoke()`

### File path variable naming

All file paths are relative, so modifiers like localFilePath are redundant. Furthermore, all paths except URL paths are file paths, so filePath is also redundant. Instead, path can be used.

### DIFF
--- a/packages/framework/src/Support/DataCollections.php
+++ b/packages/framework/src/Support/DataCollections.php
@@ -7,7 +7,6 @@ namespace Hyde\Support;
 use Hyde\Facades\Filesystem;
 use Hyde\Framework\Actions\MarkdownFileParser;
 use Hyde\Framework\Concerns\InteractsWithDirectories;
-use Hyde\Markdown\Models\MarkdownDocument;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
 use function json_decode;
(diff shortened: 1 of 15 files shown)


packages/framework/src/Support/DataCollections.php
```diff
@@ -7,6 +7,7 @@
 use Hyde\Facades\Filesystem;
 use Hyde\Framework\Actions\MarkdownFileParser;
 use Hyde\Framework\Concerns\InteractsWithDirectories;
+use Hyde\Markdown\Models\MarkdownDocument;
```
caendesilva, line 10:
```suggestion
```

Remove unused import

caendesilva (commented):
All files reviewed.

PR #1239, opened 2023-03-08
Update RealtimeCompiler to support URLs with periods

caendesilva:
Update RealtimeCompiler PageRouter to get route directly from page as route key normalization breaks when there is a period in the URL.

This is because things like `http://localhost:8080/docs/1.x/index.html` would be parsed as `http://localhost:8080/docs/1/x/index.html` which results in `Route [docs/1/x/index] not found.`

This then begs the question if we should support normalizing route keys to support dot notation at all.


caendesilva:
Closed in favour of removing support for this entirely. https://github.com/hydephp/develop/issues/1240

PR #1264, opened 2023-03-13
Use new fixes from StyleCI

caendesilva:
Apparently StyleCI has changed, which changes a large bunch of our files. Let's see what it wants this time.

### DIFF
--- a/packages/framework/src/Foundation/Kernel/PageCollection.php
+++ b/packages/framework/src/Foundation/Kernel/PageCollection.php
@@ -13,6 +13,7 @@ use Hyde\Support\Filesystem\SourceFile;
  * The PageCollection contains all the instantiated pages.
  *
  * @template T of \Hyde\Pages\Concerns\HydePage
+ *
  * @template-extends \Hyde\Foundation\Concerns\BaseFoundationCollection<string, T>
  *
  * @property array<string, HydePage> $items The pages in the collection.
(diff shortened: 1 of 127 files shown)


packages/framework/src/Foundation/Kernel/PageCollection.php
```diff
@@ -13,6 +13,7 @@
  * The PageCollection contains all the instantiated pages.
  *
  * @template T of \Hyde\Pages\Concerns\HydePage
+ *
```
caendesilva, line 16: Not as happy about this but whatever

packages/framework/src/Support/Models/RouteList.php
```diff
@@ -9,6 +9,7 @@
 
 /**
  * @internal This class is experimental and is subject to change.
+ *
```
caendesilva, line 12: Don't like this, but can probably configure

packages/framework/src/helpers.php
```diff
@@ -31,7 +31,8 @@ function unslash(string $string): string
     use Hyde\Foundation\HydeKernel;
     use Illuminate\Contracts\Support\Arrayable;
     use Symfony\Component\Yaml\Yaml;
-    use function function_exists;
+
+use function function_exists;
```
caendesilva, line 35: Not sure what happened here

PR #1285, opened 2023-03-14
Update ReadingTime to round up to one minute

caendesilva:
Fixes https://github.com/hydephp/develop/issues/1286 by rounding up the `Stringable` results of the `getFormatted` method in the `ReadingTime` class.

### DIFF
--- a/packages/framework/src/Support/ReadingTime.php
+++ b/packages/framework/src/Support/ReadingTime.php
@@ -31,6 +31,7 @@ class ReadingTime implements Stringable
 
     /** @var int The number of seconds it takes to read the text. */
     protected int $seconds;
+    protected bool $roundUp;
 
     public static function fromString(string $text): static
     {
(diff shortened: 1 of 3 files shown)


packages/framework/src/Support/ReadingTime.php
```diff
@@ -81,7 +81,7 @@ public function getSecondsOver(): int
 
     public function getFormatted(string $format = '%dmin, %dsec'): string
     {
-        return sprintf($format, $this->getMinutes(), $this->getSecondsOver());
+        return sprintf($format, $this->getMinutes() ?: 1, $this->getMinutes() < 1 ? 0 : $this->getSecondsOver());
     }
 
     /** @param  \Closure(int, int): string $closure The closure will receive the minutes and seconds as integers and should return a string. */
```
caendesilva, line 87: Am not adding any rounding to closure format as it should be as "pure" as possible since I assume anyone using this is because they want to run very custom logic.

Issue #1316, opened 2023-03-17
Bug: The blog post feed component does not work on nested pages due to not using routes


caendesilva:
### Steps to reproduce:

1.  Run the following commands:

```bash
php hyde publish:homepage posts -n
mkdir _pages/posts
mv _pages/index.blade.php _pages/posts/index.blade.php
touch _posts/foo.md
php hyde build
```

2.  Visit `_site/posts/index.html`
3.  Click on the link for the post
4.  It leads to `_site/posts/posts/foo.html` instead of `_site/posts/foo.html`

### Expected behavior:

Clicking on the link for the post should lead to `_site/posts/foo.html`.

### Actual behavior:

Clicking on the link for the post leads to `_site/posts/posts/foo.html`.

### Cause of the issue:

The issue is caused by `article-excerpt.blade.php` where the link uses the legacy style of links: `href="posts/{{ Hyde::formatLink($post->identifier . '.html') }}"` instead of `$post->getRoute`, which dynamically resolves relative paths.

### Possible solution:

Use `$post->getRoute` instead of `Hyde::formatLink` in `article-excerpt.blade.php`.

PR #1317, opened 2023-03-17
Update article excerpt component to use route helper method instead of legacy link formatter

caendesilva:
Description
-----------

This PR fixes the issue where the blog post feed component was not working on nested pages due to incorrect link formatting. The issue was caused by the use of the legacy style of links in `article-excerpt.blade.php`. The fix changes the link format to use `$post->getRoute`, which dynamically resolves relative paths.

The change was made by replacing the following code:

```blade
- <a href="posts/{{ Hyde::formatLink($post->identifier . '.html') }}"> Read post</a>
+ <a href="{{ $post->getRoute() }}" > Read post</a>
```

This PR has been tested and verified to resolve the issue. This fixes https://github.com/hydephp/develop/issues/1316


caendesilva:
I also ran a file check to ensure that we don't use the formatLink helper anywhere else in the Blade views. So this bug should be an isolated event.

PR #1335, opened 2023-03-18
Make the MetadataBag serializable

caendesilva:
Experimental for now, as I need to figure out how to serialize the metadata items without it being breaking.

While the base MetadataBag class now implements a new interface, it also adds all the needed methods, so I don't consider that change to be breaking. However since the elements are not extending a base class but instead depend on an interface I can't change that without breaking SemVer even though the impact would be minimal as the feature is undocumented and I can't imagine anyone having created custom element implementations, I am committed to not having any breaking changed between major releases.

Side effects:
- Since the class is now Jsonable, any Json representations of the kernel, it's page collection, or any individual pages, will of course include the new Json output.
- It also means that page Json uses `"metadata": []` instead of `"metadata": {}`.


caendesilva:
Think I will have to close this for now as it's a breaking change since it effects the Json schema.

PR #1339, opened 2023-03-20
Skip unit test when not running in the monorepo

caendesilva:
Fixes unrelated test failures in https://github.com/hydephp/framework/pull/572. We mostly have this check to catch any mistakes before merging config changes when developing in the monorepo.


caendesilva:
Fix is verified to be effective in https://github.com/hydephp/framework/pull/572/https://github.com/hydephp/framework/pull/572/commits/a70f3533a5b353a5e2c09a8655c917bcee7a5b00

PR #1375, opened 2023-05-22
Fix improper webpack public path to be `_site` instead of `_media`

caendesilva:
This misconfiguration is likely the cause of https://github.com/hydephp/hyde/issues/224, which this should fix.


caendesilva:
Closed in favour of https://github.com/hydephp/develop/pull/1376

PR #1376, opened 2023-05-22
Invert Webpack asset synchronization flow

caendesilva:
This method seems more reliable than existing behaviour, and should provide a better fix for https://github.com/hydephp/hyde/issues/224 than https://github.com/hydephp/develop/pull/1375.

### DIFF
--- a/webpack.mix.js
+++ b/webpack.mix.js
@@ -4,9 +4,9 @@
 
 let mix = require('laravel-mix');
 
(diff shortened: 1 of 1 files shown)


caendesilva:
System testing CI caught a side effect I missed, the mix manifest ends up in a different location.

https://github.com/hydephp/develop/actions/runs/5047110370/jobs/9053512838

PR #1380, opened 2023-09-14
Update BladeMatter documentation

caendesilva:


packages/framework/src/helpers.php
```diff
@@ -32,7 +32,7 @@ function unslash(string $string): string
     use Illuminate\Contracts\Support\Arrayable;
     use Symfony\Component\Yaml\Yaml;
 
-use function function_exists;
+    use function function_exists;
```
caendesilva, line 35: Cherrypick to main

PR #1408, opened 2023-10-28
Remove low priority todo comments

caendesilva:


packages/framework/src/Framework/Features/Blogging/Models/FeaturedImage.php
```diff
@@ -225,9 +225,7 @@ protected function getContentLengthForLocalImage(): int
 
     protected function getContentLengthForRemoteImage(): int
     {
-        // TODO: We may want to globalize this check in the config, but for now,
-        // we just check the server arguments and skip remote requests if
-        // the --no-api flag is present (in the build command call)
+        // Check if the --no-api flag is set when running the build command, and if so, skip the API call.
```
caendesilva, line 228: Wait, I realised we do have a central location in the loadRuntimeConfiguration method

caendesilva: Fixed in https://github.com/hydephp/develop/pull/1412

PR #1441, opened 2023-11-08
Update internal testing scripts

caendesilva:


caendesilva:
Apparently we can't get the job id. https://stackoverflow.com/questions/71240338/obtain-job-id-from-a-workflow-run-using-contexts

PR #1444, opened 2023-11-08
Fancy serve command

caendesilva:
image


caendesilva:
Mac screenshot for reference 
image

PR #1487, opened 2023-12-09
Fix failing browser tests

caendesilva:


caendesilva:
We'll want to fix an older version. See https://github.com/percy/cli/issues/1459#issuecomment-1850232794

caendesilva:
Fixed in upstream https://github.com/percy/cli/pull/1463

PR #1498, opened 2023-12-13
Refactor search index handling

caendesilva:


packages/framework/src/Framework/Actions/GeneratesDocumentationSearchIndex.php
```diff
@@ -48,7 +59,7 @@ protected function __construct()
     protected function run(): void
     {
         DocumentationPage::all()->each(function (DocumentationPage $page): void {
-            if (! in_array($page->identifier, Config::getArray('docs.exclude_from_search', []))) {
+            if (! in_array($page->identifier, array_merge(Config::getArray('docs.exclude_from_search', []), ['search']))) {
```
caendesilva, line 62: Could cause issues if the user wants to document their search page, maybe we tie it into the create_search_page config option? Because if they have a custom search page, they would want to disable that feature, and then this should also be disabled.

caendesilva: See 95dfaf14246d85b6de754a7618e5f1f968473c37

RELEASE_NOTES.md
```diff
@@ -10,16 +10,17 @@ This serves two purposes:
 2. At release time, you can move the Unreleased section changes into a new release version section.
 
 ### Added
-- for new features.
+- Add experimental `InMemoryPage::getOutputPath` overload support
 
 ### Changed
-- for changes in existing functionality.
+- Changed how the documentation search is generated, to be an `InMemoryPage` instead of a post-build task.
+- Updated the documentation article component to support existing `$document` instance.
```
caendesilva, line 17: May want to cherry pick this after further testing. We may also want to consider renaming the variable, but need to think about if that is a breaking change.

RELEASE_NOTES.md
```diff
@@ -10,16 +10,17 @@ This serves two purposes:
 2. At release time, you can move the Unreleased section changes into a new release version section.
 
 ### Added
-- for new features.
+- Add experimental `InMemoryPage::getOutputPath` overload support
```
caendesilva, line 13: Don't think we're using this any longer

caendesilva: Reverted in 0e757f656f56b57e1a24c5842fcdb8dccddf8b0c

RELEASE_NOTES.md
caendesilva: Since it's just a local variable inside a single Blade file I can't imagine it's very breaking, unless someone includes a child template within an extended version of our layout but (famous last words) I don't see why anyone would want to do that (and actually if they extend the template they would not get the change)

packages/framework/resources/views/components/docs/documentation-article.blade.php
```diff
@@ -1,5 +1,5 @@
 @php
-    $document = \Hyde\Framework\Features\Documentation\SemanticDocumentationArticle::make($page);
+    $document ??= \Hyde\Framework\Features\Documentation\SemanticDocumentationArticle::make($page);
```
caendesilva, line 2: Should be cherry picked, see above

RELEASE_NOTES.md
caendesilva: https://github.com/hydephp/develop/pull/1506

packages/framework/resources/views/components/docs/documentation-article.blade.php
caendesilva: Cherry picked in https://github.com/hydephp/develop/pull/1506

PR #1519, opened 2024-02-02
Make heading permalink hashes hidden when not hovered

caendesilva:
Should fix a bug where crawlers like Google Bot thinks this is a part of the heading, leading to it being visible in search results.

image

Adding `display: hidden` would likely solve this, but that breaks the transition. Using visibility modifiers should work the same, but retain the existing visual behaviour.

image


caendesilva:
Nope, this just breaks heading permalinks, and has no effect on Google indexing.

PR #1538, opened 2024-02-08
Refactor navigation internals to store navigation in the kernel

caendesilva:


caendesilva:
This breaks due to NavItem constructor forcing routes to strings which makes the links resolve too early. We can't evaluate dynamic links until individual page compile time.

packages/framework/src/Foundation/Concerns/ManagesHydeKernel.php
```diff
@@ -71,6 +72,13 @@ public function getMediaOutputDirectory(): string
         return ltrim($this->getMediaDirectory(), '_');
     }
 
+    public function navigation(): NavigationMenu
+    {
+        $this->needsToBeBooted();
```
caendesilva, line 77: This may be undesirable. It means that if someone tries to add to the menu from their code before the kernel is booted this could cause premature booting. Instead we may want to throw a helpful exception.

caendesilva:
Consider updating dropdown handling according to following proposed docs:

Dropdowns are represented by a `NavItem` with a `children` array containing more `NavItem` instances.

caendesilva:
Wondering if this API could be used in the config to initialize menus. It's kinda Filamenty.

 ```php
   'menus' => [
        NavigationMenu::create('main')
            ->withPageTypes('all', except: [DocumentationPage::class])
            ->withoutPages('404')
            ->withLabels([
                'home' => 'Start',
                'about' => 'About Us',
            ])
            ->withPriorities([
                'home' => 100,
                'about' => 200,
            ]),
        
        NavigationMenu::create('docs')
            ->withPageTypes(DocumentationPage::class)
            ->withLabels([
                'readme' => 'Readme',
                'installation' => 'Installation',
                'getting-started' => 'Getting Started',
            ])
            ->withPriorities([
                'readme' => 100,
                'installation' => 200,
                'getting-started' => 300,
            ])
            ->withoutPages([
                '404',
                'changelog'
            ]),
    ],
```

We could of course add the defaults to the base classes, so only overwrites would be needed. 

They can be accessed like this:

```php
$main = NavigationMenu::get('main'); // Main menu
$docs = NavigationMenu::get('docs'); // Sidebar
$foo = NavigationMenu::get('foo'); // Throws with message stating menu must be specified in config

```

And users could easily add their own navigation menus.

```php
    NavigationMenu::create('footer')
        ->withRoutes(routeKeys: [
            'privacy-policy',
            'terms-of-service',
        ])
        ->withNavItems([
            NavItem::create(...['some data']),
            NavItem::create(...['some data']),
        ])
        ->withLinks([
            'External Link' => 'https://example.com',
        ])
    
        // Or shorthand
        ->with(routeKeys: [
            'privacy-policy',
            'terms-of-service',
        ], links: [
            'External Link' => 'https://example.com',
        ]),
```

caendesilva:
I think I need to hold of on this implementation for a bit. It doesn't support multiple navigation menus well.

PR #1565, opened 2024-02-13
HydePHP v2.x Development

caendesilva:
# HydePHP v2.x Development 

## Abstract

This pull request tracks the code for HydePHP v2.x.

Roadmap:
- [x] https://github.com/hydephp/develop/issues/1499

## Pull Requests
- [x] https://github.com/hydephp/develop/pull/1497
- [x] https://github.com/hydephp/develop/pull/1498
- [x] https://github.com/hydephp/develop/pull/1536
- [x] https://github.com/hydephp/develop/pull/1538
- [x] https://github.com/hydephp/develop/pull/1539
- [x] https://github.com/hydephp/develop/pull/1566
- [x] https://github.com/hydephp/develop/pull/1568
- [x] https://github.com/hydephp/develop/pull/1569
- [x] https://github.com/hydephp/develop/pull/1622
- [x] https://github.com/hydephp/develop/pull/1572
- [x] https://github.com/hydephp/develop/pull/1573
- [x] https://github.com/hydephp/develop/pull/1574
- [x] https://github.com/hydephp/develop/pull/1575
- [x] https://github.com/hydephp/develop/pull/1576
- [x] https://github.com/hydephp/develop/pull/1577
- [x] https://github.com/hydephp/develop/pull/1578
- [x] https://github.com/hydephp/develop/pull/1579
- [x] https://github.com/hydephp/develop/pull/1580
- [x] https://github.com/hydephp/develop/pull/1581
- [x] https://github.com/hydephp/develop/pull/1583
- [x] https://github.com/hydephp/develop/pull/1584
- [x] https://github.com/hydephp/develop/pull/1586
- [x] https://github.com/hydephp/develop/pull/1587
- [x] https://github.com/hydephp/develop/pull/1594
- [x] https://github.com/hydephp/develop/pull/1597
- [x] https://github.com/hydephp/develop/pull/1598
- [x] https://github.com/hydephp/develop/pull/1599
- [x] https://github.com/hydephp/develop/pull/1600
- [x] https://github.com/hydephp/develop/pull/1601
- [x] https://github.com/hydephp/develop/pull/1605
- [x] https://github.com/hydephp/develop/pull/1606
- [x] https://github.com/hydephp/develop/pull/1622
- [x] https://github.com/hydephp/develop/pull/1625
- [x] https://github.com/hydephp/develop/pull/1626
- [x] https://github.com/hydephp/develop/pull/1630
- [x] https://github.com/hydephp/develop/pull/1632
[...]

Issue #1508, opened 2023-12-25
[v2.0] Refactor navigation internals


caendesilva:
For v2.0 I want to refactor navigation internals to better unify the configuration API and reduce discrepancies with the sidebar code. I think we should also consider storing the global navigation items in the Hyde kernel as that could make it easier for extensions to programmatically add navigation items.

caendesilva:
Need to consider how to handle the sidebar, but initially I'm gonna try to add the navigation as a kernel property

caendesilva:
Thinking more about this, and if the navigation belongs in the kernel. Not all sites need navigation, and it's strongly coupled to a component, so why should the kernel be responsible for booting the navigation? 

The original problem I'm trying to solve is that the only way for say a package to add a navigation menu item is by adding it to the config at runtime which feels quite hacky. This is because the navigation menu instance is created from the navigation view. 

Maybe it makes more sense to instantiate the navigation earlier on, but in a provider or similar. Then we can expose a navigation API around that. 95% of sites still probably will use navigation, so maybe we still add a helper in the kernel to access it.

caendesilva:
> From https://github.com/hydephp/develop/pull/1538#issuecomment-1936359331

Wondering if this API could be used in the config to initialize menus. It's kinda Filamenty. Since we don't cache configs (except for in the standalone which uses a caching hack) it should not be a problem. Worst case, that can be serialized in our configuration loader.

 ```php
   'menus' => [
        NavigationMenu::create('main')
            ->withPageTypes('all', except: [DocumentationPage::class])
            ->withoutPages('404')
            ->withLabels([
                'home' => 'Start',
                'about' => 'About Us',
            ])
            ->withPriorities([
                'home' => 100,
                'about' => 200,
            ]),
        
        NavigationMenu::create('docs')
            ->withPageTypes(DocumentationPage::class)
            ->withLabels([
                'readme' => 'Readme',
                'installation' => 'Installation',
                'getting-started' => 'Getting Started',
            ])
            ->withPriorities([
                'readme' => 100,
                'installation' => 200,
                'getting-started' => 300,
            ])
            ->withoutPages([
                '404',
                'changelog'
            ]),
    ],
```

They can be accessed like this:

```php
$main = NavigationMenu::get('main'); // Main menu
$docs = NavigationMenu::get('docs'); // Sidebar
$foo = NavigationMenu::get('foo'); // Throws with message stating menu must be specified in config

```

[...]

PR #1568, opened 2024-02-13
[2.x] Rewrite navigation internals

caendesilva:
TLDR: This breaks... everything. The entire Navigation API is essentially rewritten. The easiest way to upgrade is going to be reimplementing existing code from scratch.

New attempt at fixing https://github.com/hydephp/develop/issues/1508 after closing https://github.com/hydephp/develop/pull/1538. It targets HydePHP v2 https://github.com/hydephp/develop/pull/1565

With this, NavigationMenu classes are no longer responsible for generating the menus, instead the API is changed so that they are just data containers. This better follows SOLID principles, and changes the point where we generate the menus in a way that's more controllable. 

This rewrites the navigation internals in a few breaking ways. (Note that this list may be incomplete/outdated, the final version will be present in the release notes)
1. **Breaking:** Change configuration option `docs.sidebar_order` to `docs.sidebar.order` 
2. **Breaking:** Change configuration option `docs.table_of_contents` to `docs.sidebar.table_of_contents`
3. **Breaking:** The `NavItem::fromRoute()` method is now `NavItem::forRoute()`
4. Breaking: The `docs.sidebar.footer` no longer supports `true` as a value
5. Breaking: Rename class `NavigationMenu` to `MainNavigationMenu`
6. Breaking: Remove class `MainNavigationMenu` (essentially replaced/rewritten to the new `NavigationMenu` class)
7. Breaking: Remove class `BaseNavigationMenu` 
8. Breaking: Remove class `DropdownNavItem`
9. Breaking: Make `BaseNavigationMenu::$items` protected
10. Breaking: `NavItem::$destination` is now a Route instead of string (This is necessary in order to defer route resolving to compile time)
11. Breaking: Remove method `DocumentationSidebar::getGroups` 
12. Breaking: Remove method `DocumentationSidebar::getItemsInGroup` 
13. Breaking: Remove method `DocumentationSidebar::makeGroupTitle` 
14. Deprecate class `BaseNavigationMenu`
15. Deprecate class `MainNavigationMenu`
16. Deprecate class `DocumentationSidebar`
17. Deprecate class `DropdownNavItem`
18. Deprecate method `DocumentationSidebar::makeGroupTitle()`
19. Merges `DropdownNavItem` features into main `NavItem` class

Changes:
- Adds a new NavigationMenu class
- Adds a new NavigationManager class
- Adds a NavigationServiceProvider to register the manager as a singleton and register a booted callback to the kernel to generate the navigation
- Adds a new child array to `NavItem` classes to manage dropdowns
- Add a `NavItem::dropdown()` facade helper
- Make navigation item duplicate filtering case insensitive (could have side effects)
- Merges https://github.com/hydephp/develop/pull/1539
- Merges https://github.com/hydephp/develop/pull/1572
- Merges https://github.com/hydephp/develop/pull/1573
- Merges https://github.com/hydephp/develop/pull/1574
- Merges https://github.com/hydephp/develop/pull/1576
- Merges https://github.com/hydephp/develop/pull/1577
- Merges https://github.com/hydephp/develop/pull/1578
- Merges https://github.com/hydephp/develop/pull/1579
- Merges https://github.com/hydephp/develop/pull/1580
- Merges https://github.com/hydephp/develop/pull/1581
- Merges https://github.com/hydephp/develop/pull/1583
- Merges https://github.com/hydephp/develop/pull/1584
- Merges https://github.com/hydephp/develop/pull/1586
- Merges https://github.com/hydephp/develop/pull/1587
- Merges https://github.com/hydephp/develop/pull/1594
- Merges https://github.com/hydephp/develop/pull/1597
- Merges https://github.com/hydephp/develop/pull/1598
- Merges https://github.com/hydephp/develop/pull/1599
- Merges https://github.com/hydephp/develop/pull/1600
- Merges https://github.com/hydephp/develop/pull/1601
- Merges https://github.com/hydephp/develop/pull/1605
- Merges https://github.com/hydephp/develop/pull/1606
- Merges https://github.com/hydephp/develop/pull/1622
- Merges https://github.com/hydephp/develop/pull/1625
- Merges https://github.com/hydephp/develop/pull/1626
- Merges https://github.com/hydephp/develop/pull/1632
- Merges https://github.com/hydephp/develop/pull/1636
- Merges https://github.com/hydephp/develop/pull/1637
- Merges https://github.com/hydephp/develop/pull/1638
- Merges https://github.com/hydephp/develop/pull/1639
- Merges https://github.com/hydephp/develop/pull/1640
- Merges https://github.com/hydephp/develop/pull/1642

Since the navigation menu is now essentially cached in the kernel, instead of being created each time the component is rendered, the kernel must be cleared during some tests.


packages/framework/tests/Feature/HydeServiceProviderTest.php
```diff
@@ -31,6 +33,7 @@
  * @covers \Hyde\Framework\HydeServiceProvider
  * @covers \Hyde\Framework\Concerns\RegistersFileLocations
  * @covers \Hyde\Foundation\Providers\ConfigurationServiceProvider
+ * @covers \Hyde\Foundation\Providers\NavigationServiceProvider
```
caendesilva, line 36: Should be in its own test

packages/framework/src/Framework/Features/Navigation/NavItem.php
```diff
@@ -14,6 +14,8 @@
 /**
  * Abstraction for a navigation menu item. Used by the MainNavigationMenu and DocumentationSidebar classes.
  *
+ * @FIXME Before merge: Add release notes from https://github.com/hydephp/develop/pull/1539
```
caendesilva, line 17: Before merge: Add release notes from https://github.com/hydephp/develop/pull/1539

caendesilva: Resolved in https://github.com/hydephp/develop/commit/c133c82ca93a12b27954aff50a69970072edf6d0

packages/framework/src/Framework/Features/Navigation/GeneratesMainNavigationMenu.php
```diff
@@ -6,30 +6,36 @@
...
 use Illuminate\Support\Collection;
+use Hyde\Foundation\Facades\Routes;
 
 use function collect;
 
-abstract class BaseNavigationMenu
+/**
+ * @experimental This class may change significantly before its release.
+ *
+ * @todo Consider making into a service which can create the sidebar as well.
```
caendesilva, line 18: Consider making into a service which can create the sidebar as well.

packages/framework/src/Framework/Features/Navigation/NavItem.php
```diff
@@ -4,14 +4,18 @@
...
 use Hyde\Support\Models\Route;
 use Illuminate\Support\Str;
 use Stringable;
+use Hyde\Support\Models\ExternalRoute;
 
 /**
- * Abstraction for a navigation menu item. Used by the NavigationMenu and DocumentationSidebar classes.
+ * Abstraction for a navigation menu item. Used by the MainNavigationMenu and DocumentationSidebar classes.
+ *
+ * @todo Consider splitting facade methods to actual facade class.
```
caendesilva, line 18: Consider splitting facade methods to actual facade class.

packages/framework/src/Framework/Features/Navigation/NavigationMenu.php
```diff
@@ -4,71 +4,22 @@
...
-            // Create a new dropdown item containing the buffered items
-            $this->items->add(new DropdownNavItem($group, $items));
-        }
-    }
-
-    protected function canAddRoute(Route $route): bool
-    {
-        return parent::canAddRoute($route) && (! $route->getPage() instanceof DocumentationPage || $route->is(DocumentationPage::homeRouteName()));
-    }
+    /** @var \Illuminate\Support\Collection<string, \Hyde\Framework\Features\Navigation\NavItem> */
```
caendesilva, line 12: The string here has been used to determine identity to remove duplicates. We don't need that here. Duplicates are already removed in the factory, and if a user pushes items to here we have no call to remove them. We could remove this entirely, or we can solve https://github.com/hydephp/develop/commit/238dac75d43f1192079d826eafb0d7778484bb1d by using a keyed priority (we might then need a check to shift priorities in case of duplicates, which may just add more complexity...)

caendesilva: Removed in 95b395ce69a1b3e0e373285adcb64aa0f3cc57d5/088d5b5e6f55d61050890c2cbe4bab312fa62bb6

packages/framework/src/Framework/Factories/NavigationDataFactory.php
```diff
@@ -210,6 +210,7 @@ private function canUseSubdirectoryForGroups(): bool
 
     private function defaultGroup(): ?string
     {
+        // TODO: It would be better if this was null in all cases, considering 'other' is used a a sort of confusing pseudo-null
```
caendesilva, line 213: It would be better if this was null in all cases, considering 'other' is used a a sort of confusing pseudo-nul

packages/framework/tests/Feature/AutomaticNavigationConfigurationsTest.php
```diff
@@ -0,0 +1,292 @@
...
+        $this->assertMenuEquals([], [
+            new MarkdownPage('foo', ['navigation.visible' => false]),
+            new MarkdownPage('bar', ['navigation.visible' => false]),
+            new MarkdownPage('baz', ['navigation.visible' => false]),
+        ]);
+    }
+
+    public function testMainNavigationMenuWithFrontMatterGroup()
+    {
+        // TODO: For new v2 system, this should insert a root item with the group name and the children as the pages
```
caendesilva, line 128: For new v2 system, this should insert a root item with the group name and the children as the pages

caendesilva:
For dropdown setups: I think we should support index pages.

Consider this file tree:
```
_pages
├── about
│   ├── about-us.md
│   ├── contact.md
```
This should add a dropdown containing the two pages, with the dropdown title "About". If one clicks the About link, the menu should open.

Now consider this, when there is an index page, we should still support dropdowns, but clicking the actual About link, the index page should be loaded. This means that the dropdown must be active when hovering.

```
_pages
├── about
│   ├── about-us.md
│   ├── contact.md
│   └── index.md
```

.idea/php-test-framework.xml
```diff
@@ -17,6 +17,7 @@
         <cache>
           <versions>
             <info id="Local" version="9.5.21" />
+            <info id="LocalH:\monorepo\vendor\autoload.php" version="10.4.2" />
```
caendesilva, line 20: See https://youtrack.jetbrains.com/issue/WI-76125/PhpStorm-has-started-to-add-an-undesired-line-to-.idea-php-test-framework.xml-when-running-Pest-PHPUnit-tests

Must be reverted before merge. This was added in a temporary commit c639ccce8faef690d9729721d3a32315844dbc8a because I got tired of having to unstage the file which is changed each time I run the tests.

packages/framework/src/Framework/Factories/NavigationDataFactory.php
caendesilva: https://github.com/hydephp/develop/pull/1572

packages/framework/src/Framework/Features/Navigation/GeneratesMainNavigationMenu.php
caendesilva: Resolved by https://github.com/hydephp/develop/pull/1579

packages/framework/src/Framework/Features/Navigation/NavigationManager.php
```diff
@@ -0,0 +1,41 @@
...
+declare(strict_types=1);
+
+namespace Hyde\Framework\Features\Navigation;
+
+use Hyde\Hyde;
+
+/**
+ * Manages the navigation menus for the project.
+ */
+class NavigationManager
```
caendesilva, line 12: This is basically a second service container with extra steps. Maybe it's easier to just bind the menu classes directly? (Unless we want to store the menus in the kernel, which we probably don't)

caendesilva: It depends on the level of flexibility we want. Do we want each specific menu to be in its own class, or do we allow anonymous menus that are just instances of the NavigationMenu class?

caendesilva: Going to experiment with the former

caendesilva: See https://github.com/hydephp/develop/pull/1580

caendesilva: A benefit with the class based approach is that it's less ambiguous than "identifiers"
image

packages/framework/resources/views/layouts/navigation.blade.php
```diff
@@ -1,9 +1,9 @@
 @php
-    $navigation = \Hyde\Framework\Features\Navigation\NavigationMenu::create();
+    $navigation = app('navigation.main');
 @endphp
 
 <nav aria-label="Main navigation" id="main-navigation" class="flex flex-wrap items-center justify-between p-4 shadow-lg sm:shadow-xl md:shadow-none dark:bg-gray-800">
-    <div class="flex flex-grow items-center flex-shrink-0 text-gray-700 dark:text-gray-200">
+    <div id="main-navigation-brand" class="flex flex-grow items-center flex-shrink-0 text-gray-700 dark:text-gray-200">
```
caendesilva, line 6: While we're breaking stuff, maybe we should normalize these ID tags and their locations.

caendesilva:
ChatGPT response when asked to compare which is best from various perspectives: Accessing configs in views, or in helper methods in a class. (Using the new documentation sidebar system as example)



> Option B, where configuration values are accessed within methods of a class, is generally preferable from a maintenance, developer experience, and ease of customization perspective. Here's why:
> 
> Encapsulation: Option B encapsulates configuration access within the methods of the DocumentationSidebar class. This encapsulation makes it easier to manage and understand where configuration values are used and how they are manipulated. It also promotes better organization and reduces the chances of accidental modification or misuse.
> 
> Flexibility: By encapsulating configuration access within methods, Option B allows for easier customization and extension. Developers can easily modify the behavior of the DocumentationSidebar class by overriding methods or extending the class. This provides greater flexibility without directly modifying the view files, making it easier to adapt the application to changing requirements.
> 
> Readability: Accessing configuration values directly within view files (Option A) can clutter the code and reduce readability, especially in larger applications. Option B separates concerns more clearly, resulting in cleaner and more maintainable code.
> 
> Testing: Option B facilitates easier testing since logic related to configuration access is contained within methods that can be easily unit tested. This promotes better test coverage and helps ensure the reliability of the codebase.
> 
> Decoupling: By accessing configuration values through methods, Option B reduces coupling between the view layer and the underlying configuration structure. This makes it easier to change the configuration format or source without affecting the view files, promoting better separation of concerns.
> 
> Overall, Option B aligns better with principles of object-oriented design and promotes cleaner, more maintainable code compared to Option A.

PR #1571, opened 2024-02-16
Experimental headless CMS integration

caendesilva:
## Abstract

This branch experiments with integrating Decap (previously Netlify CMS) into HydePHP.

### Design & Implementation

There are a few ways we can ship the integration:
- A command that scaffolds the configuration files that are needed
  - The command can then prompt for things like GitHub repo and deployment URL
- An extension to the Realtime Compiler which serves the files needed, and can inject environment variables
- A separate Composer plugin

Currently the implementation is through a decoupled `_admin/` directory with a custom PHP server to handle dynamic injections. This can then later be adapted to the desired bundling option.

Since the CMS connects to GitHub, we could even add an option to add the static admin panel HTML and its generated config to the built site, as authentication is handled by GitHub.

We could even experiment with having a hosted version at cms.hydephp.com where users can connect any repo and we can proxy the files from there.

### Background and Stack

The resulting integration is very lightweight, as the React application is loaded from a CDN. We only need to provide the configuration data and base HTML page.

### Benefits

The main benefit of us providing this, compared to for example the user setting up their own headless CMS (Decap or otherwise), is that we our configuration file is already implemented for HydePHP sites, drastically reducing the friction.

Hopefully, something like this will make it easier for agencies and freelancers to create websites with HydePHP for their clients, who can then later update content by themselves.


_admin/config.yml
```diff
@@ -0,0 +1,59 @@
...
+    fields:
+      - { label: "Title", name: "title", widget: "string" }
+      - { label: "Cover Image", name: "image", widget: "image" }
+      - { label: "Description", name: "Description", widget: "string" }
+      - { label: "Category", name: "category", widget: "string" }
+      - { label: "Date", name: "date", widget: "datetime" }
+      - { label: "Author", name: "author", widget: "string" }
+      - { label: "Body", name: "body", widget: "markdown" }
+
+  - name: "markdown-pages"
```
caendesilva, line 37: We should also support deep navigation data and image data

caendesilva:
Since this doesn't work with Blade pages, and it doesn't seem possible/easy to implement that due to bugs in the CMS, this might not be a good option to include with HydePHP. Maybe instead we can turn it into a blog post, and make provide the default config.

caendesilva:
Closing for now, reasons noted in https://github.com/hydephp/develop/pull/1571#issuecomment-1949083783

Changes are merged into https://github.com/hyde-staging/experimental-decap-integration

caendesilva:
Might be worth looking into https://tina.io/? https://twitter.com/AnthonyCastrio/status/1788587705523228726

PR #1575, opened 2024-02-20
[2.x] Replace navigation item group normalization juggling with looser comparison 

caendesilva:
This targets v2.x via https://github.com/hydephp/develop/pull/1568

The reason we slugified the label is to have a looser comparison, considering "Foo" and "foo" the same for navigation groups. But messing with the title then makes it harder to format it back. By instead slugifying at comparison time, we bypass this problem. It also has the "side effect" of having more cases where the title is accurate, hence why some tests needed updating.

This is obviously susceptible to race conditions, but there is only so much we can do if the user is not consistent with their group naming. It also kinda makes the internal state "ugly" but since it reduces complexity and improves accuracy I think it's worth it.


caendesilva:
Realizing the part of the problem I'm facing is that the generator doesn't now about all the context when assembling dropdowns. Maybe we need an accessor instead (or add a second factory pass)

PR #1588, opened 2024-02-24
Improved view testing

caendesilva:
Adds some testing helpers and a new test. This PR is considered internal since it only affects the test suite and no public code.


caendesilva:
Realized/remembered that Laravel brings view testing features. I don't think they're 100% the best, so am going to change this to make some improvements to the existing ones.

PR #1592, opened 2024-02-25
Improved helper functions

caendesilva:
Adds helper function documentation and adds a few new global helper functions:

- Add a global `asset()` helper function
- Add a global `route()` helper function
- Add a global `url()` helper function

Design notes:

I debated weather these functions should be global or under the Hyde function namespace. I think making them global is the best. For the majority of users, this will be the most convenient, and for a small subset of users where there are conflicts (such as including Hyde in a Laravel application (which is not "officially" supported)), we can add a constant to disable these functions in case of function registration race conditions. TLDR: I think the small conflict risk is worth the major convenience of global functions. 

Compatibility mode:

To disable the new global functions, define a constant called `HYDE_COMPATIBILITY_MODE` and set it to `true`, and the new functions will not be registered. The constant should be defined early on, for example in a bootstrap file.


caendesilva:
Reminder to people reading this: `HYDE_COMPATIBILITY_MODE` **must** be defined **before** the autoloader is required.

PR #1601, opened 2024-03-01
[2.x] Extract back NavItem class for dropdowns

caendesilva:
Adds back the dedicated class for grouped navigation items, but implemented better this time. This better separates concerns, and makes the data state of both classes clearer, and reduces the need for conditional returns and type logic.


caendesilva:
Could also draw inspiration from Filament of their `->items` method
https://github.com/filamentphp/filament/blob/3.x/packages/panels/src/Navigation/NavigationGroup.php#L65-L73

PR #1610, opened 2024-03-15
Clean up HTML and Blade files

caendesilva:


caendesilva:
Think https://github.com/hydephp/develop/edit/clean-up-html-and-blade/packages/framework/resources/views/components/article-excerpt.blade.php needs some formatting

PR #1617, opened 2024-03-17
Internal: Move base testing classes to testing package

caendesilva:
This essentially reverts the change in https://github.com/hydephp/develop/pull/1607, as it for some reason breaks framework tests on macOS. Instead, these base classes are now autoloaded using the testing package.


caendesilva:
Initial Ubuntu test failed, likely due to race condition as it was triggered before the monorepo package was split and propagated to the testing package.

PR #1630, opened 2024-03-22
[2.x] Extract trait for shared code with navigation menu class

caendesilva:


caendesilva:
The implementations are quite different actually, with the grouped navigation items being stored as an array. Refactoring this isn't worth the complexity, instead I think it's better to simplify the API of the grouped item class.

caendesilva:
Reopening trying to target https://github.com/hydephp/develop/pull/1642

caendesilva:
Partially reverted by https://github.com/hydephp/develop/pull/1645

PR #1635, opened 2024-03-24
[2.x] Experiment with splitting out the route model

caendesilva:
Targets https://github.com/hydephp/develop/pull/1568 and experiments with extracting a PageRoute model.

This may add too much complexity, and not solve the right problem (as the issue is with ExternalRoutes being hacky, but since they are new and just used in navigation it may be better to improve that instead), but I wanted to visualize what this could look like before making a decision.


caendesilva:
Yeah, no. This is definitely not solving the right problem.

PR #1638, opened 2024-03-24
[2.x] Create better fallbacks when not supplying a navigation item label

caendesilva:
Targets https://github.com/hydephp/develop/pull/1568

Since null is not allowed, we needed to coalesce to an empty string. This makes no sense from a practical usability perspective: What's the point of a link in the menu that you can't click because there is no label to click? It makes more sense to fall back to the link to make it usable, and give a nudge to the developer that they may want to change it into something more visually pleasing.

### DIFF
--- a/packages/framework/src/Framework/Features/Navigation/NavigationItem.php
+++ b/packages/framework/src/Framework/Features/Navigation/NavigationItem.php
@@ -68,7 +68,7 @@ class NavigationItem implements NavigationElement, Stringable
             $group ??= $destination->getPage()->navigationMenuGroup();
         }
 
-        return new static($destination, $label ?? $destination, $priority ?? NavigationMenu::DEFAULT, $group);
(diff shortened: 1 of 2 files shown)


packages/framework/src/Framework/Features/Navigation/NavigationItem.php
```diff
@@ -137,4 +137,14 @@ public static function normalizeGroupKey(?string $group): ?string
     {
         return $group ? Str::slug($group) : null;
     }
+
+    /** @experimental This feature may be removed before release, and the label will just fall back to the input $url */
+    protected static function makeTitleFromUrl(string $url): string
```
caendesilva, line 142: This is probably going too far. I don't want to encourage people using direct links this much.

caendesilva: Reverted in https://github.com/hydephp/develop/commit/8f2bd586a87f3e29a2c24978d53996b50d7655bd

PR #1642, opened 2024-03-26
[2.x] Simplify and normalize public navigation APIs

caendesilva:
Targets https://github.com/hydephp/develop/pull/1568 as I when writing the docs realized that having two `NavigationItem` constructors with differing syntaxes and internal logic adds a ton of complexity for no real reason. This PR <s>protects the constructor to force a single unified API through the static `create` method.</s> unifies the constructor signatures and logic to provide the exact same results regardless of the syntax used.

It also changes navigation groups to use collections instead of an array for the internal state to match the navigation group system.


caendesilva:
Consider reopening https://github.com/hydephp/develop/pull/1630

PR #1651, opened 2024-04-09
[2.x] Simplify the Feature enum class

caendesilva:


caendesilva:
All these patches were essentially applied to master and merged into the target branch, hence why the file diff is now empty.

PR #1657, opened 2024-04-12
Internal: Try to fix Codecov

caendesilva:
Tries to fix Codecov, again.


caendesilva:
Codecov really seems to be broken. They want me to use a token, but I can't add a token because it refuses to let me log in as the organisation even though they have all the permissions they need. Reinstalling the GitHub App does nothing for them to realise that I should be able to manage the organisation on Codecov.

caendesilva:
Reopening as they now seem to have given me access to the org

PR #1660, opened 2024-04-13
Skip sitemap generation when there is no valid base URL

caendesilva:
Before, this would fall back to a relative URL, but these no longer validate by Google and will not be indexed. It feels a bit dramatic to have an exception be thrown when building test sites locally so instead we just skip the build, but with a helpful message.

image

Technically this can be seen as a breaking change, as this results in a sitemap file not being generated. But the end result in both cases is the same. Before this change we in this scenario when missing a base URL we get a sitemap that cannot be indexed, and after this change we don't generate a sitemap at all. In both cases, Google will not index the site using the sitemap. In the new case, we at least provide a notice to the user. I think this better follows the principle of least astonishment and thus I believe this justifies targeting HydePHP v1.6 instead of v2.0.

The reason we consider localhost to be a non-valid URL is a tradeoff between users wanting to see what the sitemap looks locally, and when building for production in a CI where they forgot to configure an environment variable. I think the latter weights more heavily here. We could also add a `--force` flag to the `build:sitemap` command, but I don't think the complexity is warranted.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -18,6 +18,7 @@ This serves two purposes:
 
 ### Changed
 - The `features` array in the `config/hyde.php` configuration file is now an array of `Feature` enums in https://github.com/hydephp/develop/pull/1650
+- Sitemap generation will now be skipped if a base URL is not set, as Google now will not index sitemaps without a base URL in https://github.com/hydephp/develop/pull/1660
 
 ### Deprecated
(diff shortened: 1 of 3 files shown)


caendesilva:
### ChatGPT code review


> Since the current behavior results in an invalid sitemap that Google won't index anyway, skipping sitemap generation when there's no valid base URL seems like a sensible change. It doesn't break any existing functionality since the previous behavior also didn't achieve the desired outcome. However, it's important to communicate this change clearly to users who might rely on the previous behavior. You could mention it in release notes or update documentation to ensure users understand why sitemap generation might be skipped in certain cases. This way, they can adjust their workflows accordingly.

PR #1663, opened 2024-04-14
Clean up and refactor code

caendesilva:
Minor code cleanup and refactors, mainly for tests. Goes through almost all tests to ensure they are properly formatted and use the best assertions. Some of these are over two years old (AKA the first tests of HydePHP) and in need of some care. It also fixes over 30 Psalm issues and increases the type coverage to as much as we're gonna be able to get it.

### DIFF
--- a/docs/architecture-concepts/the-hydekernel.md
+++ b/docs/architecture-concepts/the-hydekernel.md
@@ -41,7 +41,7 @@ app(HydeKernel::class)->version();
 hyde()->version();
 ```
 
-The Kernel instance is constructed in `bootstrap.php`, and is available globally as `$hyde`.
+The Kernel instance is constructed and bound in the `app/bootstrap.php` file.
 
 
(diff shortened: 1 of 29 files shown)


caendesilva:
Since the assertEquals method in PHPUnit is final, I hacked the vendor code to add this to the top of the method to detect scalar usages. Noting here in case I need it again.

```php
        $isScalarArray = function($array) {
            if (!is_array($array)) {
                return false;
            }

            foreach ($array as $value) {
                if (!is_scalar($value)) {
                    return false;
                }
            }

            return true;
        };

        if ((is_scalar($expected) && is_scalar($actual)) || ($isScalarArray($expected) && $isScalarArray($actual))) {
            echo(sprintf("Expectation with scalar values should use assertSame on line %s:%s",
                debug_backtrace()[0]['file'],
                debug_backtrace()[0]['line'])
            );
        }

```

Issue #1523, opened 2024-02-06
Configuration loader should be able to set documentation sidebar title from Hyde YAML config


caendesilva:
```yml
# hyde.yml
name: Example
```

This should also affect the sidebar header

caendesilva:
How I wish we could reference values in other config files. The following also has the same issue

```php
Meta::property('site_name', env('SITE_NAME', 'HydePHP')),
```

PR #1691, opened 2024-04-26
Configuration loader improvements 

caendesilva:


caendesilva:
I can't come up with a good way to fix https://github.com/hydephp/develop/issues/1523 without adding a ton of complexity to handle all edge cases. I think https://github.com/hydephp/develop/pull/1692 is much more viable for this.

PR #1697, opened 2024-04-28
Add a config option to disable theme toggle buttons to automatically use browser settings

caendesilva:
You can now disable the theme toggle buttons by setting `theme_toggle_buttons` to `false` in `config/hyde.php`.

If the `Feature::Darkmode` setting is disabled in the `features` array in the same file, this won't do anything, but if darkmode is enabled setting the theme toggle buttons setting to false they will not show up in the app layout nor the documentation layout; instead the appropriate color scheme will be automatically applied based on the browser system settings.

### DIFF
--- a/packages/framework/src/Facades/Features.php
+++ b/packages/framework/src/Facades/Features.php
@@ -89,6 +89,11 @@ class Features implements SerializableContract
         return static::enabled(Feature::Darkmode);
     }
 
+    public static function hasDarkmodeButtons(): bool
+    {
+        return static::hasDarkmode() && Config::getBool('hyde.darkmode_buttons', true);
+    }
+
     /**
      * Torchlight is by default enabled automatically when an API token
      * is set in the .env file but is disabled when running tests.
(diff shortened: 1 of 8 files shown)


packages/framework/src/Facades/Features.php
```diff
@@ -89,6 +89,11 @@ public static function hasDarkmode(): bool
         return static::enabled(Feature::Darkmode);
     }
 
+    public static function hasDarkmodeButtons(): bool
```
caendesilva, line 92: Might be better to be `hasThemeToggle`?

packages/framework/src/Facades/Features.php
```diff
@@ -89,6 +89,11 @@ public static function hasDarkmode(): bool
...
+    {
+        return static::hasDarkmode() && Config::getBool('hyde.darkmode_buttons', true);
+    }
```
caendesilva, line 95: `theme_toggle_buttons` or `theme_toggle`?

caendesilva:
Note that if this is enabled for an existing site, if a user has previously used the toggle buttons they may still have a local storage setting that will take precedence. Clearing this would require changing the storage key (which we may want to do in a future release to prefix it with `hyde`) but we can't do that here as it's a breaking change.

PR #1700, opened 2024-04-29
.gitattributes: substitute CRLF only in text files

ydalton:
Changing CRLF to LF makes sense for text files, but not images, where doing so can lead to file corruption. This commit tells Git to ignore binary files and only do the substitution in plain text files.

Fixes hydephp/hyde#239.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
-* text eol=lf
+* text=auto eol=lf
 
 *.blade.php diff=html
 *.css diff=css


caendesilva:
Okay, double checked with ChatGPT and it seems like I have indeed been at fault here. Thank you so much for making this PR! I'm gonna go ahead and merge this, and I'll also go over all the gitattribute files in the organization to make sure we do this properly.

> The difference lies in how Git determines whether a file should be treated as text and have its line endings normalized to LF.
> 
> 1. `* text eol=lf`: This rule explicitly tells Git that all files should be treated as text files, regardless of their content. It applies LF line endings to all files, irrespective of whether Git considers them as text or binary.
> 
> 2. `* text=auto eol=lf`: This rule tells Git to automatically detect whether a file is text or binary. If Git determines a file to be text (based on its content), it will apply LF line endings to it. If Git determines a file to be binary, it will not modify its line endings.
> 
> In summary, the difference lies in the specificity of the rule:
> 
> - With `text`, all files are explicitly treated as text.
> - With `text=auto`, Git automatically determines whether a file is text or binary.

ydalton:
Thanks for the merge. Regardless, I did testing as you said...

Okay, so I created a new project with `composer create-project hyde/hyde` and created a Git repository, got a PNG file, then as a control I converted `_pages/index.blade.php` to a CRLF file using `unix2dos`, and added those to Git, where I got this:
```
pino@my-distrobox:~/Documents/Projects/hyde$ git add .
warning: in the working copy of '_media/header.png', CRLF will be replaced by LF the next time Git touches it
warning: in the working copy of '_pages/index.blade.php', CRLF will be replaced by LF the next time Git touches it
```
I then changed the `.gitattributes` like in this commit, and added the files the same way, now with a better looking output:
```
pino@my-distrobox:~/Documents/Projects/hyde$ git add .
warning: in the working copy of '_pages/index.blade.php', CRLF will be replaced by LF the next time Git touche
s it
```
By the way, thank you for this lovely static site generator!

caendesilva:
Thank you so much for testing it!

> I then changed the `.gitattributes` like in this commit, and added the files the same way, now with a better looking output:
> 
> ```
> pino@my-distrobox:~/Documents/Projects/hyde$ git add .
> warning: in the working copy of '_pages/index.blade.php', CRLF will be replaced by LF the next time Git touche
> s it
> ```

Awesome, that's just what I want. Since I do a lot of my development on a Windows machine I want to normalize text files (like Blade files for example) but I obviously don't want that to happen to binary files.


> By the way, thank you for this lovely static site generator!

Thank you for the kind words and all your help debugging and patching this issue!

Issue #1515, opened 2024-01-30
When setting nagivation groups for standard pages in front matter the dropdown subdirectory setting is required


caendesilva:
Using the following in `_pages/page.md`
```markdown
---
navigation.group: My Group
---
```
Has no effect unless the following is set in `config/hyde.php`
```php
'navigation' => [
    'subdirectories' => 'dropdown',
]
```

This feels wrong, since this does not necessarily mean the user wants to use subdirectories.

PR #1704, opened 2024-04-30
[2.x] Honour front matter navigation groups set in front matter regardless of subdirectory setting

caendesilva:
Merges https://github.com/hydephp/develop/pull/1703 into https://github.com/hydephp/develop/pull/1565


caendesilva:
Actually, it seems like we had already fixed https://github.com/hydephp/develop/issues/1515 in the v2 branch, so rescoping this PR to just include the new tests to prove the fix.

PR #1710, opened 2024-05-04
[2.x] Improve the numerical filename priority prefix feature

caendesilva:
Refactors some code introduced in https://github.com/hydephp/develop/pull/1709 and implements the remaining documented tasks and adds a lot more tests as this is an important feature that is implemented through several interconnected parts of the codebase, thus unit tests are added to make sure the code units operate as expected; as mainly high level feature tests were added in the original PR.


packages/framework/tests/Unit/RouteKeyTest.php
```diff
@@ -93,6 +95,56 @@ public function testFromPageWithCustomNestedOutputDirectory()
...
+        $this->assertSame('foo', RouteKey::fromPage(InMemoryPage::class, '00123-foo')->get());
+        $this->assertSame('foo', RouteKey::fromPage(InMemoryPage::class, '000123-foo')->get());
+        $this->assertSame('foo', RouteKey::fromPage(InMemoryPage::class, '0000123-foo')->get());
+    }
+
+    public function testItExtractsCoreIdentifierPartFromNumericalFilenamePrefixForNestedIdentifiers()
+    {
+        $this->assertSame('foo/bar', RouteKey::fromPage(InMemoryPage::class, 'foo/01-bar')->get());
+        $this->assertSame('foo/bar/baz', RouteKey::fromPage(InMemoryPage::class, 'foo/bar/02-baz')->get());
+        $this->assertSame('foo/01-bar/baz', RouteKey::fromPage(InMemoryPage::class, 'foo/01-bar/03-baz')->get());
```
caendesilva, line 130: We need to see how we should handle this. Presumably we would want to remove the prefix from subdirectories too, since we have documented that it works for subdirectories (which still needs to be implemented). This could get complex...

caendesilva: Resolved in 0bcb133786efff896138ea203188c17caedbabc3

Issue #1720, opened 2024-06-09
Ensure a better state when a site URL is not set


caendesilva:
We need to handle cases where a site URL is not set better. For example, deploying using the Action and a hyde.yml file, the following gets added:

```
<link rel="canonical" href="http://localhost">
<link rel="sitemap" href="http://localhost/sitemap.xml" type="application/xml" title="Sitemap">
```

I don't think this is ever desirable, as localhost URLs will never be valid in production, and locally aa relative URL will still be resolved.

So, I think we need to add broad level tests to ensure no localhosts get added to builds.

Additionally, we need relative links to sitemaps, and we should only add canonical tags when they are valid - and localhost are not.

PR #1726, opened 2024-06-12
Improve behaviour when a site URL is not explicitly set

caendesilva:
## Abstract

We're getting some issues by Hyde using `localhost` as the default site URL. Initially, this PR will target the 1.x branch because I consider these things to be bugs that should be fixed in v1, however, there may be side effects of this: primarily that output HTML will change - but I do not think anyone will depend on those undesired effects before this patch, since localhost is never desired in production. This will resolve https://github.com/hydephp/develop/issues/1720

Expanding on the reasoning behind allowing this in 1.x: Since v1 is LTS, I don't want to leave those users hanging with what is in my opinion a bug, that localhost links can "leak" through production. And I think the only way to fix this is by changing some logic which could have side effects. But I think I am okay with that tradeoff as it stands now, especially since I think any possible usages that may be affected probably don't actually desire the logic we have now.

### Method `Hyde::hasSiteUrl()` now returns false if URL is localhost

This better matches the documented method description of "Check if a site base URL has been set in config (or .env).", as this value is `localhost` by default and thus not set by the user.

### Method `Hyde::url()` now resolves relative paths instead of throwing

We now return the relative path even if the base URL is not set, as this is more likely to be the desired behavior the user's expecting. If the user is trying to get the base URL, but it's not set, we throw the exception.

### DIFF
--- a/packages/framework/src/Foundation/Kernel/Hyperlinks.php
+++ b/packages/framework/src/Foundation/Kernel/Hyperlinks.php
@@ -131,7 +131,7 @@ class Hyperlinks
     {
         $value = Config::getNullableString('hyde.url');
 
-        return ! blank($value);
+        return ! blank($value) && $value !== 'http://localhost';
     }
 
     /**
--- a/packages/framework/src/Foundation/Kernel/Hyperlinks.php
+++ b/packages/framework/src/Foundation/Kernel/Hyperlinks.php
@@ -129,7 +129,9 @@ class Hyperlinks
      */
     public function hasSiteUrl(): bool
     {
-        return ! blank(Config::getNullableString('hyde.url'));
+        $value = Config::getNullableString('hyde.url');
+
+        return ! blank($value);
     }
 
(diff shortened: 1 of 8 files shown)


caendesilva:
We can probably update code added in https://github.com/hydephp/develop/pull/1660 to use the new logic

packages/framework/src/Foundation/Kernel/Hyperlinks.php
```diff
@@ -147,6 +149,13 @@ public function url(string $path = ''): string
...
         }
 
+        // Since v1.7.0, we return the relative path even if the base URL is not set,
+        // as this is more likely to be the desired behavior the user's expecting.
+        if (! blank($path)) {
+            return $path;
+        }
+
+        // User is trying to get the base URL, but it's not set
         throw new BaseUrlNotSetException();
```
caendesilva, line 159: We might consider this for deprecation

Issue #1712, opened 2024-06-05
[2.x] DataCollections improvements


caendesilva:
## Overview
Changes needed to improve the data collections feature, that we save for v2 as they may be breaking

### Support standalone Yaml files

We need to change the parsing so we don't need front matter blocks. (Fixed in https://github.com/hydephp/develop/pull/1733)

https://github.com/hydephp/develop/blob/4af2234e89deb567bceaaeaba8230d8c1109ee7c/docs/digging-deeper/collections.md?plain=1#L137

### Normalize class name

It bugs me that it is DataCollections instead of DataCollection. This works fine when it's a facade, so `DataCollections::get()` is semantic* (so is `DataCollection::get()`), however, since these classes are _also_ used as instances containing data, it makes less sense for it to be plural, because now calling `DataCollections::markdown()` implies (many) DataCollections due to the return. Singularizing this would also match the Illuminate Collection.

Additionally, no Laravel facades are plural, however our `Includes` is plural (and some others, but Includes is a sibling class to this and should match)

*`DataCollections::get()` is not fully semantic as it may imply it will return multiple collections.

caendesilva:
We may also want to consider excluding README.md files from Markdown collections

caendesilva:
Should we add a warning if a data collection file is empty?

caendesilva:
Fixed in https://github.com/hydephp/develop/pull/1732

PR #1732, opened 2024-06-23
[2.x] DataCollections improvements

caendesilva:
See the 1.x tie-in to make the upgrade smoother https://github.com/hydephp/develop/pull/1733. Fixes https://github.com/hydephp/develop/issues/1712

Main changes:
- **Breaking:** Renamed class `DataCollections` to `DataCollection`
    - Fixes part of https://github.com/hydephp/develop/issues/1712
- JSON collection files are now validated
    - Fixes https://github.com/hydephp/develop/issues/1736
- Yaml data files no longer need to start with triple dashes to be parsed by DataCollections
    - Fixes part of https://github.com/hydephp/develop/issues/1712
    - (cherry picked to 1.x via https://github.com/hydephp/develop/pull/1733)
- No longer automatically create the data collections directory 
    - Fixes https://github.com/hydephp/develop/issues/1735 
- Adds a new custom `ParseException` exception

Also adds increased unit testing for syntax options and fault tolerance

### DIFF
--- a/packages/framework/src/Support/DataCollection.php
+++ b/packages/framework/src/Support/DataCollection.php
@@ -6,23 +6,23 @@ namespace Hyde\Support;
 
 use stdClass;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Str;
 use Hyde\Facades\Filesystem;
 use Symfony\Component\Yaml\Yaml;
+use Illuminate\Support\Collection;
 use Hyde\Markdown\Models\FrontMatter;
 use Hyde\Markdown\Models\MarkdownDocument;
 use Hyde\Framework\Actions\MarkdownFileParser;
 use Symfony\Component\Yaml\Exception\ParseException;
-use Illuminate\Support\Collection;
-use Illuminate\Support\Str;
 
 use function blank;
 use function rtrim;
 use function implode;
 use function explode;
 use function ucfirst;
-use function json_decode;
 use function sprintf;
 use function unslash;
+use function json_decode;
 use function str_replace;
 use function json_last_error_msg;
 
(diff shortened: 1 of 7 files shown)


caendesilva:
Questioning if we should support `---` at all, but they are valid (and recommended) spec so I can't justify the BC break of removing them. Still, only the first Yaml document in a file is valid, anything else will be trimmed, and I don't think it's common enough to try this to warrant an exception here. We do handle it gracefully by using `Str::between` (originally implemented to simplify logic to trim front matter encased Yaml with dashes both above and under)

caendesilva:
Technically we don't gain any new information from having these full array keys, and calling them "identifiers" internally makes it seem like they would match the page identifiers.

The current:

```php
$collection = \Hyde\Support\DataCollection::markdown('testimonials');

\Hyde\Support\DataCollection {
    "testimonials/1.md" => Hyde\Markdown\Models\MarkdownDocument
    "testimonials/2.md" => Hyde\Markdown\Models\MarkdownDocument
    "testimonials/3.md" => Hyde\Markdown\Models\MarkdownDocument
  ]
}
```

Could just as well be this:
```php
$collection = \Hyde\Support\DataCollection::markdown('testimonials');

\Hyde\Support\DataCollection {
    "1" => Hyde\Markdown\Models\MarkdownDocument
    "2" => Hyde\Markdown\Models\MarkdownDocument
    "3" => Hyde\Markdown\Models\MarkdownDocument
  ]
}
```

If you call a Markdown collection you know the file extension, and you just provided the subdirectory name. I'm not sure it's worth the BC though...

packages/framework/src/Support/DataCollection.php
caendesilva, line 1: Can be reverted if we implement https://github.com/hydephp/develop/issues/1736

packages/framework/src/Support/DataCollection.php
```diff
@@ -45,49 +48,75 @@ class DataCollections extends Collection
...
 
-        return new static(static::findFiles($name, ['yaml', 'yml'])->mapWithKeys(function (string $file): array {
-            return [static::makeIdentifier($file) => MarkdownFileParser::parse($file)->matter()];
-        }));
+            $content = Str::between($content, '---', '---');
+
+            try {
+                $parsed = Yaml::parse($content) ?: [];
+            } catch (ParseException $exception) {
+                throw new InvalidArgumentException(sprintf("Invalid YAML in file: '%s' (%s)", $file, $exception->getMessage()), previous: $exception);
```
caendesilva, line 77: Was wondering if we need to add the message from the parse error here, and I think we do, because the previous exception does show up in the realtime compiler, but not when running the build command

image




image

Issue #1734, opened 2024-06-23
RFC: Should Includes helpers returning Html return HtmlStrings?


caendesilva:
We have this class, where we document that "Both Markdown and Blade includes will be rendered to HTML". Wouldn't it then make more sense if these returned `HtmlString` objects instead of `string` types?

```php
/**
 * The Includes facade provides a simple way to access partials in the includes directory.
 *
 * Both Markdown and Blade includes will be rendered to HTML.
 */
class Includes
```

This is more semantic, but has the side effect that they will no longer be escaped. 

Considering the usages though, I imagine that these are already being used within literal echo directives `{!!  !!}`. It could be a breaking change, and it needs to be communicated clearly, so possibly something for v2.x?

caendesilva:
##  GPT-4o Response

### Feedback on RFC: Should Includes Helpers Returning Html Return HtmlStrings?

#### Summary

The proposal suggests modifying the return type of HTML content from plain strings to `HtmlString` objects in the `Includes` class of the HydePHP static site generator. This change aims to improve semantic correctness by explicitly indicating that the returned content is HTML and should not be escaped.

#### Points in Favor

1. **Semantic Clarity**:
   - Returning `HtmlString` instead of plain strings makes the intent explicit. It clearly indicates that the content is HTML and should be handled appropriately by the rendering system.

2. **Preventing Double Escaping**:
   - By returning `HtmlString`, you avoid the risk of double escaping HTML content. This ensures that the content is rendered correctly when included in views.

3. **Alignment with Documentation**:
   - The current documentation states that "Both Markdown and Blade includes will be rendered to HTML". Using `HtmlString` aligns the implementation with this documentation, providing consistency and clarity.

#### Points of Concern

1. **Breaking Changes**:
   - Changing the return type to `HtmlString` can break existing implementations that expect a plain string. This requires clear communication and possibly a major version update (e.g., v2.x).

2. **Security Implications**:
   - HtmlStrings are not escaped by default, which can introduce security risks if the content is not properly sanitized. This change requires careful consideration of how content is sourced and processed to avoid XSS vulnerabilities.

3. **Usage Considerations**:
   - The current implementation might be relying on `{!! !!}` directives to prevent escaping. While changing to `HtmlString` could simplify usage, it might require developers to review and update existing code to ensure compatibility.

#### Recommendations

1. **Versioning and Communication**:
[...]

PR #1738, opened 2024-06-26
[2.x] Includes facade improvements

caendesilva:
- [x] Methods in the `Includes` facade now return `HtmlString` objects instead of `string`
    - Fixes https://github.com/hydephp/develop/issues/1734
- [x] Markdown includes are now converted to HTML using the custom HydePHP Markdown service, meaning they now support full GFM spec and custom Hyde features like colored blockquotes and code block filepath labels 
- [x] Markdown returned from includes are now trimmed of trailing whitespace and newlines 
- [x] The `Includes::path()` and `Includes::get()` methods now normalizes paths to be basenames to match the behaviour of the other include method.
    - This means that nested directories are no longer supported, as you should use a data collection for that.

### Includes facade changes

The following methods in the `Includes` facade now return `HtmlString` objects instead of `string`:

- `Includes::html()`
- `Includes::blade()`
- `Includes::markdown()`

Notes:

- This means that you no longer need to use `{!! !!}` to render the output of these methods in Blade templates, instead just use `{{ }}`.
- If you have used the return value of these methods in custom code, you may need to adjust your code to work with the new return type.

For more information, see the RFC that proposed this change: https://github.com/hydephp/develop/issues/1734
The RFC was implemented in https://github.com/hydephp/develop/pull/1738

#### Remember to escape output if necessary

**Note:** Remember that this means that includes are **no longer escaped** by default, so make sure to escape the output if necessary, for example if the content is user-generated.
- (Use `{{ e(Includes::html('foo')) }}` instead of `{{ Includes::html('foo') }}` to escape the output, matching the previous behavior.)

### DIFF
--- a/packages/framework/src/Support/Includes.php
+++ b/packages/framework/src/Support/Includes.php
@@ -90,11 +90,13 @@ class Includes
             if ($default === null) {
                 return null;
             } else {
-                return new HtmlString(trim(Markdown::render($default, MarkdownDocument::class)));
+                $markdown = $default;
+                return new HtmlString(trim(Markdown::render($markdown, MarkdownDocument::class)));
             }
         }
 
-        return new HtmlString(trim(Markdown::render(Filesystem::get($path), MarkdownDocument::class)));
+        $markdown = Filesystem::get($path);
+        return new HtmlString(trim(Markdown::render($markdown, MarkdownDocument::class)));
     }
 
     /**
--- a/packages/framework/src/Support/Includes.php
+++ b/packages/framework/src/Support/Includes.php
@@ -91,7 +91,6 @@ class Includes
                 return null;
             } else {
                 $markdown = $default;
-                return new HtmlString(trim(Markdown::render($markdown, MarkdownDocument::class)));
             }
         } else {
             $markdown = Filesystem::get($path);
--- a/packages/framework/src/Support/Includes.php
+++ b/packages/framework/src/Support/Includes.php
@@ -87,7 +87,7 @@ class Includes
         $path = static::normalizePath($filename, '.md');
         $contents = static::getFileContents($path);
 
-        if (! Filesystem::exists($path)) {
+        if ($contents === null) {
             return $default === null ? null : static::renderMarkdown($default);
         }
 
--- a/packages/framework/src/Support/Includes.php
+++ b/packages/framework/src/Support/Includes.php
@@ -87,16 +87,10 @@ class Includes
         $path = static::normalizePath($filename, '.md');
 
         if (! Filesystem::exists($path)) {
-            if ($default === null) {
-                return null;
-            } else {
-                $contents = null;
-            }
-        } else {
-            $contents = Filesystem::get($path);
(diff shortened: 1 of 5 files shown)


caendesilva:
Could we hook into the Blade `@include` directive? (And should we?)

packages/framework/src/Support/Includes.php
```diff
@@ -84,10 +85,10 @@ public static function markdown(string $filename, ?string $default = null): ?str
         $path = static::normalizePath($filename, '.md');
 
         if (! Filesystem::exists($path)) {
-            return $default === null ? null : Markdown::render($default);
+            return $default === null ? null : Markdown::render($default, HydePage::class);
         }
 
-        return Markdown::render(Filesystem::get($path));
+        return Markdown::render(Filesystem::get($path), HydePage::class);
```
caendesilva, line 91: Considered if we should try to inherit the rendered page's class here, but don't think it would offer value. I'm also thinking that we may actually want to have some different logic here (maybe passing the base Markdown document class) since we probably do not want Torchlight attributions injected into partials, as they could get repeated a lot.

caendesilva: Resolved in f15bea8eb839e0121b166a9ed30c6b89811d0104 and 83c613649bcb2199a452251254e2d17479ed81ed

caendesilva:
Somewhat related/general tip: Setting the includes path in `view.php` allows Blade partials from the same directory to be accessed using the `@include` directive

packages/framework/tests/Feature/IncludesFacadeTest.php
```diff
@@ -113,4 +116,139 @@ public function testBladeReturnsRenderedDefaultValueWhenNotFound()
...
+    public function testTorchlightAttributionIsNotInjectedToMarkdownPartials()
+    {
+        $this->file('resources/includes/without-torchlight.md', 'Syntax highlighted by torchlight.dev');
+
+        $this->assertSame(
+            '<p>Syntax highlighted by torchlight.dev</p>',
+            Includes::markdown('without-torchlight.md')
+        );
+    }
+
```
caendesilva, line 129: This is obviously not working, did something wrong here. Also: Document in release notes

caendesilva: Nvm, that's just to trigger the thing. Gonna clarify it.

packages/framework/src/Support/Includes.php
```diff
@@ -42,70 +44,75 @@ public static function path(?string $filename = null): string
...
+            return $default === null ? null : new HtmlString(Blade::render($default));
         }
 
-        return Blade::render(file_get_contents($path));
+        return new HtmlString(Blade::render(Filesystem::get($path)));
+    }
+
+    protected static function normalizePath(string $filename, string $extension): string
+    {
+        return static::path(basename($filename, $extension).$extension);
```
caendesilva, line 116: Thinking if we should support using subdirectories (could be done by removing `basename`), but I think that might be an anti-pattern, as if you need grouped data you probably want to use a data collection instead.

PR #1770, opened 2024-06-30
Improve the Yaml configuration loader

caendesilva:
**Overview:**
- Refactors and improves the code
- New attempt at https://github.com/hydephp/develop/pull/1691
- Fixes https://github.com/hydephp/develop/issues/1523

### DIFF
--- a/packages/framework/src/Foundation/Internal/LoadYamlConfiguration.php
+++ b/packages/framework/src/Foundation/Internal/LoadYamlConfiguration.php
@@ -17,6 +17,8 @@ use function file_exists;
 /**
  * @internal Bootstrap service that loads the YAML configuration file.
  *
+ * @implements \LaravelZero\Framework\Contracts\BoostrapperContract [sic]
+ *
  * @see docs/digging-deeper/customization.md#yaml-configuration
  *
  * It also supports loading multiple configuration namespaces, where a configuration namespace is defined
--- a/packages/framework/src/Foundation/Internal/LoadYamlConfiguration.php
+++ b/packages/framework/src/Foundation/Internal/LoadYamlConfiguration.php
@@ -44,6 +44,8 @@ class LoadYamlConfiguration
             $this->config = Config::all();
             $this->yaml = $this->parseYamlFile();
 
+            $this->supportSettingSiteNameSetsSidebarHeaderOption();
+
             $this->mergeParsedConfiguration();
 
             Config::set($this->config);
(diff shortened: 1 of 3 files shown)


caendesilva:
We _could_ also check environment values for the site name, but I think that adds to much complexity.

packages/framework/src/Foundation/Internal/LoadYamlConfiguration.php
```diff
@@ -80,14 +92,26 @@ protected function mergeParsedConfiguration(): void
...
 
     protected function configurationContainsNamespaces(array $yaml): bool
     {
         return array_key_first($yaml) === 'hyde';
     }
+
+    private function supportSettingSiteNameSetsSidebarHeaderOption(): void
+    {
+        $sidebarHeaderIsNotSetInPhpConfig = ($this->config['docs']['sidebar']['header'] ?? null) === 'HydePHP Docs';
+        $siteNameFromYaml = $this->configurationContainsNamespaces($this->yaml) ? ($this->yaml['hyde']['name'] ?? null) : ($this->yaml['name'] ?? null);
```
caendesilva, line 109: We could also check environment values for the site name, but I think that adds to much complexity.

caendesilva: Might be worth to at least try to implement this and see how complex it is, seeing as it would mean the `env-site-name` setting in the https://github.com/hydephp/action would then work for this automatically.

caendesilva: Question is the, if it's set in both Yaml and .env, which takes precedence? I think .env as that usually has precedence generally speaking

caendesilva: Wait, I was wrong again. This should already be implicitly supported by the very fact of an environment variable being used. Originally I considered if we should do things the other way around but that's why we closed https://github.com/hydephp/develop/pull/1691

caendesilva:
Unless this is modified to another pattern, we should also support this

        'description' => env('SITE_NAME', 'HydePHP').' RSS Feed',


And of course this

        Meta::property('site_name', env('SITE_NAME', 'HydePHP')),

I think these are all the remaining ones

caendesilva:
It's a bit hard to test, but I confirmed manually that setting the site name in the environment variables, that takes precedence over all this, as those are evaluated before this.

caendesilva:
> ```
>     Meta::property('site_name', env('SITE_NAME', 'HydePHP')),
> ```

Not sure how we can set this, since it's not keyed.

PR #1771, opened 2024-07-01
Improve Yaml environment configuration loading

caendesilva:
Picks up where https://github.com/hydephp/develop/pull/1770 left off

> I think adding a dedicated environment configuration loader to get data from the Yaml file could work if it's in between the environment and configuration loaders


caendesilva:
Okay, GitHub glitched and merged this when merging the cherry picked PR...

PR #1773, opened 2024-07-01
Refactor bootstrappers and improve Yaml environment configuration loading

caendesilva:
Picks up where https://github.com/hydephp/develop/pull/1770 left off and replaces https://github.com/hydephp/develop/pull/1771 which was merged by GitHub due to a [glitch](https://github.com/hydephp/develop/pull/1771#issuecomment-2200297519).

> I think adding a dedicated environment configuration loader to get data from the Yaml file could work if it's in between the environment and configuration loaders

This pull request is a story of how I continuously lost my mind.

In short, what we ended up with is a simple system that reads the yaml configuration before the application boots, and injects an environment variable for the site name if it's needed in order to elegantly have it be used in the configuration which is loaded afterwards.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -22,7 +22,7 @@ This serves two purposes:
 - Method `Hyde::url()` will now return a relative URL instead of throwing an exception when supplied a path even if the site URL is not set in https://github.com/hydephp/develop/pull/1726
 - Updated the `.env.example` file to contain more details on the site URL setting's usages in https://github.com/hydephp/develop/pull/1746
 - Added a version prefix to the sitemap's generator attribute in https://github.com/hydephp/develop/pull/1767
-- Setting a site name in the Yaml config file will now be used to set the documentation sidebar header in https://github.com/hydephp/develop/pull/1770
+- Setting a site name in the Yaml config file will now influence all configuration values where this is used, unless already set, in https://github.com/hydephp/develop/pull/1770 and https://github.com/hydephp/develop/pull/1773
 
 ### Deprecated
 - Deprecated the global `unslash()` function, replaced with the existing namespaced `\Hyde\unslash()` function in https://github.com/hydephp/develop/pull/1753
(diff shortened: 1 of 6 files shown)


caendesilva:
Okay, I confirmed that at this point this is working with real usage, but due to the dump way the Laravel environment system is designed, it's very hard to test this.

caendesilva:
At this point it feels much easier to just introduce a new syntax:

  ```php
  'meta' => [
      Meta::name('generator', 'HydePHP v'.Hyde\Hyde::version()),
        Meta::property('site_name', '{{ $site_name }}'),
    ],

```

**Comment by caendesilva** [comment]
Ahhhhhh somewhere we lost the original config repository instance

```php
array:1 [
  "set repository data for class instance id" => 940
]
array:1 [
  "set repository data for class instance id" => 1237
]
array:1 [
  "got repository data for class instance id" => 940
]
```

caendesilva:
> At this point it feels much easier to just introduce a new syntax:
> 
> ```
> 'meta' => [
>     Meta::name('generator', 'HydePHP v'.Hyde\Hyde::version()),
>       Meta::property('site_name', '{{ $site_name }}'),
>   ],
> ```
> 
> This could easily then be injected using a config post-processor

Yeah, I give up. This is what we're doing:

```
'meta' => [
    Meta::property('site_name', '{{ env.site_name }}'),
],
```

This way, it's impossible for the parser to confuse it for Blade syntax, but anyone familiar with templating can tell it will be substituted, and the `.env` prefix less magic where it comes from.

caendesilva:
Considering how much we need to fight three frameworks here due to them hardcoding environment variables, there is really no easy way to implement this without having to reinvent a whole lot of wheels. Most of the code I made work when running globally, but there is no good way to reset things between test cases.

caendesilva:
Given how small feature this technically is, maybe we can go without having it tested for now. Reverting to last stable state

caendesilva:
State of current manual testing:

If env is set and site name in yml config is set: env takes precedence
If site name in yml config is set: value is used

Both cases, the value is used for both doc headers, feeds and meta

And neither one of these methods are called (but the faulty tests depend on them)

            $this->supportSettingSidebarHeaderFromSiteName();
            $this->supportSettingRssFeedTitleFromSiteName();

caendesilva:
It would simplify things if we could add a custom Env class, but for that we need to add a custom `env` function, but the one from Laravel takes priority...

PR #1793, opened 2024-07-06
Extract helper method for automatic serialization

caendesilva:
Since we can't do things like `parent::toArray` to utilize the trait logic, and we can't do `$this->toArray` from an overridden method, this allows classes using this to work on the original logic like this:

```php
use Serializable;

public function toArray(): array
{
    return array_filter($this->automaticallySerialize());
}
```

### DIFF
--- a/packages/framework/src/Support/Concerns/Serializable.php
+++ b/packages/framework/src/Support/Concerns/Serializable.php
@@ -17,9 +17,7 @@ trait Serializable
     /** Default implementation to dynamically serialize all public properties. Can be overridden for increased control. */
     public function toArray(): array
     {
-        // Calling the function from a different scope means we only get the public properties.
-
-        return get_object_vars(...)->__invoke($this);
+        return $this->automaticallySerialize();
     }
 
     /** Recursively serialize Arrayables */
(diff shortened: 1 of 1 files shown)


caendesilva:
### Claude response on method naming:

The added method is currently named `automaticallySerialize()`. While this name gives an idea of what the method does, it could be improved to be more specific and aligned with existing naming conventions. Here are a few suggestions:

1. `getSerializableArray()`: This name clearly indicates that the method returns an array suitable for serialization.

2. `getOriginalArrayRepresentation()`: This emphasizes that it's returning the original array representation before any filtering or modification.

3. `getBaseSerializableData()`: This suggests that it's providing the base data for serialization, which can be further processed.

4. `toArrayInternal()`: This name follows the convention of the existing `toArray()` method but indicates it's an internal implementation.

5. `getSerializableProperties()`: This name focuses on the fact that it's returning the serializable properties of the object.

Of these options, I would recommend `getSerializableArray()` or `toArrayInternal()`. These names are concise, clear, and follow common naming conventions in PHP.

The choice between them depends on whether you want to emphasize the "get" aspect (`getSerializableArray()`) or align more closely with the existing `toArray()` method (`toArrayInternal()`).

Using one of these improved names, the example usage would look like this:

```php
use Serializable;
public function toArray(): array
{
    return array_filter($this->getSerializableArray());
}
```

or

```php
use Serializable;
public function toArray(): array
{
    return array_filter($this->toArrayInternal());
}
```

These names more accurately reflect the purpose of the method and its relationship to the serialization process.

caendesilva:
I think `automaticallySerialize` is better than the Claude suggestions.

Issue #1762, opened 2024-06-30
Generate author pages


caendesilva:
I think it's about time we generate dynamic pages for authors and blog post categories. We can have a feature flag system to control which of these are created, as well as an easy way to support setting the templates. We could also provide built-in pagination support for blog posts as a part of this rework.

See sister issue https://github.com/hydephp/develop/issues/1831 which tracks category pages

caendesilva:
Maybe we do this through extensions?

caendesilva:
Splitting this out so that we have one issue for authors and one for navigation links

caendesilva:
We will have to reconsider where the author's name should link to: (their website or profile page?)

![image](https://github.com/user-attachments/assets/ce518252-42e8-4d0b-838e-55538f2659a1)

caendesilva:
These are essentially just https://github.com/hydephp/publications, but backed by dynamic data (authors/tags), wondering if we should not provide this as a prebuilt publication type thing instead.

PR #1798, opened 2024-07-07
[2.x] Breaking: Improve the new Authors API

caendesilva:


packages/framework/src/Framework/Features/Blogging/Models/PostAuthor.php
```diff
@@ -87,26 +86,38 @@ public function __construct(string $username, ?string $name = null, ?string $web
         $this->socials = $socials;
     }
 
+    /**
+     * Create a new Post Author instance from an array of data.
+     *
+     * @param  array{username?: string, name?: string, website?: string, bio?: string, avatar?: string, socials?: array<string, string>}  $data
+     */
+    public static function create(array $data): PostAuthor
```
caendesilva, line 94: Check if we should use `create` or `make` to match our conventions. Using `make` may be beneficial to differentiate from the `Author` facade method with very different semantics.

packages/framework/tests/Unit/PostAuthorTest.php
```diff
@@ -203,57 +258,107 @@ public function testAllMethodReturnsEmptyCollectionIfNoAuthorsAreSetInConfig()
...
         $this->assertInstanceOf(PostAuthor::class, $author);
         $this->assertSame('foo', $author->username);
         $this->assertSame('bar', $author->name);
     }
 
-    public function testGetMethodReturnsNewAuthorIfUsernameNotFoundInConfig()
+    public function testGetMethodNormalizesUsernamesForRetrieval()
+    {
+        Config::set('hyde.authors', [
+            'foo_bar' => Author::create(),
```
caendesilva, line 302: Should we normalize 'foo bar' to 'foo_bar' when loading the configuration?

However, this might not be necessary because it should match the array key, and normalizing that could be unexpected.

caendesilva: Claude:

Consider normalizing usernames in the `Config::set()` call:

The current implementation uses `foo_bar` as the key in the `hyde.authors` configuration. However, the test suggests that various formats (e.g., 'foo bar', 'Foo Bar', 'FOO BAR') should be accepted. 

There are two potential approaches to address this:

1. Normalize the config key:
   Pros: Consistent with retrieval behavior.
   Cons: May lead to unexpected behavior if users rely on exact key matching.

2. Keep the config key as-is and normalize only during retrieval:
   Pros: Maintains expected behavior for config keys.
   Cons: Introduces a discrepancy between config keys and retrieval methods.

Recommendation: Clarify the intended behavior and document the chosen approach. If normalization is desired, consider implementing it consistently across both configuration and retrieval.

Additionally, update the test description to reflect the final decision on normalization behavior.

caendesilva: I think after considering the Claude notes, we may want to go with normalization across the board. We normalize for a reason, and it's weird to support the user going against that. If they know what they are doing, they can modify the Author class.

caendesilva: A normalization would mean that the following results in a single author:

```php
Config::set('hyde.authors', [
    'foo_bar' => Author::create(),
    'foo-bar' => Author::create(),
    'foo bar' => Author::create(),
    'Foo Bar' => Author::create(),
    'FOO BAR' => Author::create(),
]);
```

But I cannot come up with a reason where such a thing would be desired.

caendesilva: **Claude**

Let's think through this:

In most real-world scenarios, it's highly unlikely that a user would intentionally want to create multiple distinct authors with names that differ only in formatting. This kind of duplication would typically be considered undesirable for several reasons:

1. Confusion: It would be confusing for content creators and readers to have multiple author profiles that appear to represent the same person.

2. Inconsistency: It could lead to inconsistent attribution of content, where the same person's work might be associated with different author profiles.

3. Data management: It would complicate data management and queries, potentially leading to incomplete or inaccurate results when trying to gather all content by a specific author.

4. SEO considerations: For websites, having multiple author pages for what is essentially the same person could dilute SEO value and potentially be seen as duplicate content.

However, there might be some edge cases where similar names could be legitimately different:

1. Localization: In some cultures, name variants might be more significant. For example, "Björn" and "Bjorn" might be considered distinct names in some contexts.

2. Pseudonyms: An author might use slightly different pen names for different genres or publications.

3. Historical figures: In academic or historical contexts, there might be a need to distinguish between different spellings of a name used in different periods or documents.

Given these considerations, it seems that normalizing usernames would be the most sensible approach for the vast majority of use cases. It would prevent accidental duplication and ensure consistency.

If you do implement normalization, it would be wise to:

1. Document the normalization rules clearly.
2. Provide a way to override normalization in the rare cases where distinct, similarly-named authors are needed.
3. Implement normalization consistently across all parts of the system (configuration, retrieval, display, etc.).

caendesilva: We could add a validator that simply checks the count of the parsed authors and the raw array and warns if they do not match.

caendesilva: Implemented in https://github.com/hydephp/develop/pull/1798/commits/d9ceeefc3b9b94e21eae40c8b3f4fac6e839be99

packages/framework/tests/Unit/PostAuthorTest.php
```diff
@@ -326,6 +326,36 @@ public function testGetMethodNormalizesUsernamesForRetrieval()
...
+            'FOO BAR' => Author::create(),
+        ]);
+
+        $this->assertCount(1, PostAuthor::all());
+    }
+
+    public function testOnlyLastAuthorWithNormalizedUsernameIsKept()
+    {
+        // Technically, a first come, first-served principle would make most sense,
+        // but this is the lightest code implementation and can be used for now.
```
caendesilva, line 345: Given the context of your application and the goal of preventing unintended duplicates, it makes the most sense to keep the first user in the array matching the normalized username. Here's why:

First-come, first-served principle: This approach respects the order in which authors were added to the configuration. The first occurrence of a normalized username is treated as the "canonical" version.
Predictability: It's easier for users to understand and predict the behavior if they know that the first occurrence will be used.
Backward compatibility: If users have existing configurations, keeping the first occurrence ensures that the system continues to use the author data they originally intended.
Performance: Checking only for the first occurrence can be slightly more efficient than scanning the entire array to find the last occurrence.

caendesilva: Implemented in a03541585e646144cf45f47b82ee2f170e6f9f9a

config/hyde.php
```diff
@@ -277,17 +277,26 @@
...
+            website: 'https://hydephp.com', // Optional website URL
+
+            // The following settings are not used in the bundled templates,
+            // but you can use them in your own custom views, for example.
+            // bio: 'The mysterious author of HydePHP',
+            // avatar: 'avatar.png',
+            // socials: [
+            //     'twitter' => 'HydeFramework',
+            //     'github' => 'hydephp',
+            // ],
```
caendesilva, line 299: Maybe we do not need this here, as they are not yet used, and they are in the documentation. However, if https://github.com/hydephp/develop/issues/1762 will use these they can be kept, hence why I'm holding off on removing them.

PR #1814, opened 2024-07-09
Internal: Version control the monorepo pages configuration

caendesilva:
Having this on another branch does not work anymore when new major versions have BC breaks, so we need to version control these alongside the code.


caendesilva:
We may also want to consider adding the fixtures not in this repository to version control, and to matrix test them.

PR #1843, opened 2024-07-12
[2.x] Dynamic blogging related page generation

caendesilva:


caendesilva:
Had an idea that we could improve the realtime compiler route not found exception handling for pages like these, by showing the conditions needed for dynamic pages to show up. For example: Tip: An automatic Authors listing page is added when you have at least one blog post with an author defined

image

caendesilva:
Some tweaks left https://search.google.com/test/rich-results/result/r%2Farticles?id=1SFc2gCwPWVFI29jqdBFCQ

image

caendesilva:
Another dynamic page could be to add a posts page.

We also need to make sure we don't add these if the pages exist.

caendesilva:
Test fixture patch https://github.com/hydephp/develop/commit/560a514d89eea90966c83c8ccd7431d1dee1ee45.patch

packages/framework/tests/Feature/DynamicAuthorPagesTest.php
```diff
@@ -0,0 +1,151 @@
...
+            '_posts/hyde_post_1.md',
+            '_posts/hyde_post_2.md',
+            '_posts/hyde_post_3.md',
+            '_posts/jane_post_1.md',
+            '_posts/jane_post_2.md',
+            'authors/index',
+            'authors/mr_hyde',
+            'authors/jane_doe',
+            // 'authors/user123', // This user has no posts, so no author page should be created
+            // TODO: 'authors/anonymous', // This will be supported as the author has made a post, and thus exists in the system, even if not in the config
```
caendesilva, line 45: Need to consider how important this is. It's gonna add a lot of complexity, either we need to find the authors beforehand (which did not go great https://github.com/hydephp/develop/pull/1846 and https://github.com/hydephp/develop/pull/1845), or we need to add them in the listing page (https://github.com/hydephp/develop/pull/1845#issuecomment-2227006201) but this means we also need to create those listing pages. Maybe we'll hold off on this until the future.

PR #1844, opened 2024-07-13
[2.x] Breaking: Update in memory pages always have the same route key as their identifier

caendesilva:
Currently a class like this:

```php
class InMemoryPageWithCustomOutputDirectory extends InMemoryPage
{
    public static string $outputDirectory = 'foo';
}
```
Will when having an identifier of "bar" have the route key "foo/bar", but the identifier "bar", which can make it conflict with other pages, even where not intended.

### DIFF
--- a/packages/framework/src/Pages/InMemoryPage.php
+++ b/packages/framework/src/Pages/InMemoryPage.php
@@ -6,13 +6,13 @@ namespace Hyde\Pages;
 
 use BadMethodCallException;
 use Closure;
-use Hyde\Support\Models\RouteKey;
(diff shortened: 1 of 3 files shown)


caendesilva:
Think we need to reverse this, so that is solves the right problem, by setting identifiers to route keys.

caendesilva:
It might be more breaking if people depend on identifiers, but I think a better way might be to solve the root of this: Make identifiers for InMemoryPages unique by adding a virtual source directory.

caendesilva:
No, I think it's better to note that developers need to be careful about this. Not worth adding a bunch of weird code to try to fix a bad idea.

PR #1845, opened 2024-07-13
[2.x] Generate authors found in blog posts

caendesilva:


caendesilva:
Okay this kinda adds a bunch of complexity, but realized the only place I currently care about this is so unconfigured authors can show up in author listing https://github.com/hydephp/develop/pull/1843, so we can just handle it there.

PR #1846, opened 2024-07-13
[2.x] Discover authors when booting the kernel

caendesilva:


caendesilva:
Wanted this for https://github.com/hydephp/develop/pull/1845, but won't work since the page factories look to this data. We would have to merge them in after discovery.

PR #1858, opened 2024-07-16
Refactor realtime compiler to handle virtual routes through the service container

caendesilva:
Creates a new system to instead of hardcoding virtual routes, we store them in the service container (which is loaded when the router boots)

This means it's also possible for package developers to add realtime compiler routes:

```php
<?php

namespace App\Providers;

use Illuminate\Support\ServiceProvider;
use Hyde\RealtimeCompiler\RealtimeCompiler;

class AppServiceProvider extends ServiceProvider
{
    /**
     * Bootstrap any application services.
     */
    public function boot(): void
    {
        $router = $this->app->make(RealtimeCompiler::class);

        $router->registerVirtualRoute('/foo', callable(\Desilva\Microserve\Request): \Desilva\Microserve\Response);
    }
}

```

### DIFF
--- a/.github/workflows/smoke-tests.yml
+++ b/.github/workflows/smoke-tests.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: vendor
-          key: ${{ runner.os }}-php-${{ hashFiles('**/composer.lock') }}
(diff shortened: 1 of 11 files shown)


caendesilva:
Rebased to target master branch

PR #1868, opened 2024-07-17
Refactor code to increase code quality and type coverage

caendesilva:
Resolves ~200 Psalm issues and brings type coverage from `97.82%` to `99.97%`


caendesilva:
Okay, literally impossible for us to get higher type coverage! Only two mixed types left, and those are developer hooks where we actually cannot know the types.

image

Issue #1774, opened 2024-07-01
Support setting root page meta description in front matter


caendesilva:
Would improve SEO by making it easier to add custom page descriptions

https://developers.google.com/search/docs/appearance/snippet#use-quality-descriptions

```php
interface PageSchema extends FrontMatterSchema
{
    public const PAGE_SCHEMA = [
        'title' => 'string',
        'canonicalUrl' => 'string', // While not present in the page data, it is supported as a front matter key for the accessor data source.
        'description' => 'string', // For <meta name='description'>.
        'navigation' => NavigationSchema::NAVIGATION_SCHEMA,
    ];
}
```

caendesilva:
## Claude naming response

Based on the context you've provided and the HydePHP philosophy, I would recommend using `description` rather than `pageDescription` for the following reasons:

1. Simplicity: The motto "Simplicity first" suggests that we should opt for the simpler, more straightforward option when possible. `description` is shorter and clearer.

2. Consistency: In the `BlogPostSchema`, you're already using `description`. Using the same term across different schemas promotes consistency and reduces cognitive load for developers.

3. Intuitive API: The philosophy emphasizes that "Code should make sense and be intuitive, especially user-facing APIs." `description` is a widely understood term in the context of meta tags and SEO, making it intuitive for both experienced and novice developers.

4. Convention over configuration: Using `description` aligns with common conventions in web development and SEO practices.

5. Familiarity: For users coming from Laravel or other PHP frameworks, `description` is a more common term for meta descriptions.

6. Avoiding redundancy: Since this is already in a `PageSchema` interface, adding "page" to the property name would be redundant.

7. Flexibility: Using the more general term `description` allows for potential future use beyond just page descriptions if needed.

While `pageDescription` might be more explicit, the context of its use within a `PageSchema` already provides sufficient clarity. The simpler `description` aligns better with HydePHP's philosophy and goals.

If you feel that additional clarity is needed, you could consider adding a comment to the property, like so:

```php
'description' => 'string', // Meta description for the page, used in <meta name="description">
```

This approach maintains simplicity in the API while providing extra context for developers who might need it.

PR #1884, opened 2024-07-23
Support setting page descriptions in front matter for all page types

caendesilva:
Fixes https://github.com/hydephp/develop/issues/1774


caendesilva:
Noticed we don't have any place in the docs where this would fit. Something for later.

PR #1590, opened 2024-02-25
[2.x] Dynamic Markdown links

caendesilva:
## PR Abstract

Based on [RFC 1590](https://rfc.hydephp.com/rfc/1590.html) proposed in https://twitter.com/CodeWithCaen/status/1761810371063497098

## Overview

HydePHP provides a powerful feature for automatically converting Markdown links to source files to the corresponding routes in the built site.

This allows for much better writing experience when using an IDE, as you can easily navigate to the source file by clicking on the link.

## Usage

Using the feature is simple: Just use source file paths for links:

```markdown
[Home](/_pages/index.blade.php)
![Logo](/_media/logo.svg)
```

As you can see, it works for both pages and media assets. The leading slash is optional and will be ignored by Hyde, but including it often gives better IDE support.

### Behind the Scenes

During the build process, HydePHP converts source paths to their corresponding routes, and evaluates them depending on the page being rendered.

If your page is in the site root then:

- `/_pages/index.blade.php` becomes `index.html`
- `/_media/logo.svg` becomes `media/logo.svg`

If your page is in a subdirectory then:

- `/_pages/index.blade.php` becomes `../index.html`
- `/_media/logo.svg` becomes `../media/logo.svg`

Of course, if your page is in a more deeply nested directory, the number of `../` will increase accordingly.

We will of course also match your configured preference for `pretty_urls` and only include the `.html` extension when desired.

### Limitations

There are some limitations and considerations to keep in mind when using this feature:

- This feature won't work for dynamic routes (not backed by a file)
- If you rename a file, links will break. Your IDE may warn about this.
- If a file is not found, we won't be able to see it when evaluating links.
- Relative links are not supported (so ../_pages/index.blade.php won't work)

PR #1901, opened 2024-07-24
[2.x] Improved dynamic Markdown links

caendesilva:
### This is part of something bigger.

- This is part of https://github.com/hydephp/develop/pull/1590
- Solves code review https://github.com/hydephp/develop/issues/1590#issuecomment-2248722665
            
Please see https://github.com/hydephp/develop/pull/1590 for information.


packages/framework/src/Markdown/Processing/DynamicMarkdownLinkProcessor.php
```diff
@@ -5,38 +5,68 @@
...
+        /** @var \Hyde\Support\Models\Route $route */
+        foreach (Hyde::routes() as $route) {
+            $map[$route->getSourcePath()] = $route;
+        }
+
+        return $map;
+    }
+
+    /**
+     * @todo Consider adding caching, since this is not a singleton like the routes.
```
caendesilva, line 53: Yeah this needs caching. Benchmark shows a constant time of about `2ms` per iteration. 

```php
// Ensure the classes are loaded before the benchmark.
MediaFile::all();

\Illuminate\Support\Benchmark::dd(function () {
    MediaFile::all();
}, 1000); // "1.884ms", "1.884ms", "2.006ms"
```

I think we can use a simple static property/variable. In real usages this won't change during the application build cycle. In unit tests we can mock the method.

caendesilva: Adding a static variable brings this so essentially zero (0.000ms)

```php
// Ensure the classes are loaded before the benchmark.
static::assetMap();

\Illuminate\Support\Benchmark::dd(function () {
    static::assetMap();
}, 100000); // 0.000ms
```

PR #1904, opened 2024-07-25
[2.x] Rewrite the Asset API

caendesilva:
### Background

**The Problem:** The current system is confusing and unintuitive with multiple and inconsistent ways to access assets.
**The Cause:** This is due to multiple code locations have had overlapping code independently implemented, without thinking about the bigger picture.
**The Solution:** This PR rewrites the web of related code into a unified Asset API, designed to provide a consistent interface to interact with the project's media asset.

### Scope

We create the new solution by thinking about how users will actually use these features, which can be summarized into the following scope:
- End users will mainly interact with Media asset files from their pages. Commonly, this will be to do things like resolving links to images located in their media folder. Given our simplicity-first approach, this is our primary focus: to provide a convenient way to get links to assets.
- A secondary role of the API is for the internal HydePHP codebase, package developers, and serious power users, to be used in parts of the static site generation to interact with the site assets.

### Podcast

Overwhelmed by this massive changelog? Listen to this AI-generated podcast episode by NotebookLM by Google which was fed the commit logs. Improved version [available here](https://cdn-static.hydephp.com/experiments/ai/podcast/deep-dive-new-asset-api-commit-log-second-version.mp3)

https://github.com/user-attachments/assets/dbb6e271-692e-44d1-a280-be564473fdb5

## Roadmap

- [x] See if we can cache the media assets during discovery. A good location may be in the kernel Filesystem instance. This would also provide a small performance optimization if we create cache busting keys at discovery time. It would also improve DX with almost no performance penalty as we could throw an exception when trying to get a non-existent file.
- [x] Throw exceptions when trying to get an `asset()` that does not exist
- [x] Handle side effect in https://github.com/hydephp/develop/pull/1911#discussion_r1693243854
- [x] Remove the `mediaLink` helper as it can be merged into the main `asset()` helper
- [x] Rename `getContentLength()` to `getLength()`
- [x] More or less implements https://github.com/hydephp/develop/issues/1981

## Development

- [x] https://github.com/hydephp/develop/pull/1907
- [x] https://github.com/hydephp/develop/pull/1908
- [x] https://github.com/hydephp/develop/pull/1909
- [x] https://github.com/hydephp/develop/pull/1910
- [x] https://github.com/hydephp/develop/pull/1911
- [x] https://github.com/hydephp/develop/pull/1913
- [x] https://github.com/hydephp/develop/pull/1914
- [x] https://github.com/hydephp/develop/pull/1915
- [x] https://github.com/hydephp/develop/pull/1917
[...]

PR #1911, opened 2024-07-26
[2.x] Normalize media path helpers to unify naming with HydePages

caendesilva:
This PR is part of https://github.com/hydephp/develop/pull/1904

```php
// Renames to match MediaFile with HydePage base
Hyde::mediaPath => MediaFile::sourcePath 
Hyde::siteMediaPath => MediaFile::outputPath 
```


caendesilva:
Thinking if we should move the directory helpers as well, but not sure if it is worth it at this time.

packages/framework/src/Foundation/Concerns/ForwardsFilesystem.php
```diff
@@ -28,21 +28,11 @@ public function vendorPath(string $path = '', string $package = 'framework'): st
...
     public function sitePath(string $path = ''): string
     {
         return $this->filesystem->sitePath($path);
     }
 
-    public function siteMediaPath(string $path = ''): string
-    {
-        return $this->filesystem->siteMediaPath($path);
-    }
-
```
caendesilva, line 45: This means they are lost from the HydeKernel which could be a problem to review before merge.

caendesilva: Think we can revert the deprecated compat trait, as root deprecations should propagate after https://github.com/hydephp/develop/pull/1919

caendesilva: Don't think this is an issue anymore.

PR #1914, opened 2024-07-26
[2.x] Remove the `mediaLink` method

caendesilva:
Part of https://github.com/hydephp/develop/pull/1904 and removes the `mediaLink` helper as it can be merged into the main asset() helper.


caendesilva:
I'm starting to remember why we have two methods. Media link was intended to be the main way to use get assets, as it uses the more compatible relative links. The second parameter was for things where specs required absolute urls like metadata. But I did not document this so I forgot it.

caendesilva:
Drafting until I consider what solves the right problems the best way.

caendesilva:
> Drafting until I consider what solves the right problems the best way.

Maybe we should improve the object API by adding `path()` and `link()` instance helpers.

caendesilva:
Superseded by https://github.com/hydephp/develop/pull/1932

PR #1917, opened 2024-07-27
[2.x] Bring media assets into the Hyde kernel

caendesilva:
This is part of https://github.com/hydephp/develop/pull/1904

>  See if we can cache the media assets during discovery. A good location may be in the kernel Filesystem instance. This would also provide a small performance optimization if we create cache busting keys at discovery time. It would also improve DX with almost no performance penalty as we could throw an exception when trying to get a non-existent file.

Quick benchmarking shows this now goes from `~2.5ms` per `MediaFile::files()` call to `~2.7ms` for the first call, and essentially zero at all subsequent calls in the request lifecycle. This is at the cost of the kernel needing to rerun discovery if media files change during the request, which should only happen in unit testing that are coupled to the filesystem.


caendesilva:
Okay, added a more robust benchmarker in , with about 15 files, including nested ones, and a 10MB one.

```yaml
# Before:
Warmup: 15.8036ms
Benchmark: 10.2105987ms # avg/1000/its

# After:
Warmup: 16.9699ms
Benchmark: 0.00093500000000001ms # avg/1000/its
```

PR #1918, opened 2024-07-28
[2.x] Major performance and data handling improvements to media assets

caendesilva:
This is part of https://github.com/hydephp/develop/pull/1904 and follows up upon https://github.com/hydephp/develop/pull/1917/

Main changes:
- MediaFile instances now have their `$path` property normalized to start with the media directory prefix, to better match the documented state in the parent class.
- Uses `CRC32` for media asset hashing (and thus cache busting) instead of `MD5` in dca20a021d646a11cbd4b79e13c697cd85c67b5f and 4cb592f5cf0ec80f47c1b2c20a10355616d85280
- Media files are now validated to exist upon constructing the instances.
- The API is designed to match HydePages, for example `files()` returning an array, `all()` returning a collection
- All data in the media file instance is computed at construct, improving performance from `O(k)` to `O(n * m)` (or `O(n)` if `m` is constant) - where `k` is the number of media files, `m` is the average number of assets included on a page, and `n` is the number of pages. [^1] This is a significant improvement, especially for large sites. [^2]

[^1]: https://github.com/hydephp/develop/pull/1918#issuecomment-2255507224
[^2]: For sites with a massive amount of media assets, this could lead to a performance decrease when rebuilding a single page / viewing a page in the realtime compiler. If this proves an issue, we can lazy load the discovery in such cases. (~150 files with ~50 MB takes less than 40ms)<br>For smaller sites the overhead is less than 8ms, so this should not be that big of an issue, see [1918#discussion_r1694291464](https://github.com/hydephp/develop/pull/1918#discussion_r1694291464)


packages/framework/src/Support/Filesystem/MediaFile.php
```diff
@@ -119,6 +128,11 @@ public function getMimeType(): string
         return 'text/plain';
     }
 
+    public function getHash(): string
+    {
+        return md5_file($this->getAbsolutePath());
```
caendesilva, line 133: Thinking about which algoritm we should use. I think `MD5` is sufficient. We don't need anything cryptographically secure. MD5 is fast, and it's a short string too.

caendesilva: ### Claude response when asked a neutral question:

For a MediaFile class in a PHP project based on Laravel, using a checksum hash for cache busting is a good approach. When choosing an algorithm for this purpose, you'll want to consider factors like speed, collision resistance, and compatibility. Here are some options, along with my recommendation:

1. MD5: Fast, but not cryptographically secure. Still widely used for non-security critical applications like cache busting.

2. SHA-1: Slightly slower than MD5, but more collision-resistant. Also commonly used for non-security purposes.

3. SHA-256: More secure, but slower than MD5 and SHA-1. Overkill for cache busting.

4. CRC32: Very fast, but has a higher collision rate. Suitable for cache busting if speed is crucial.

For your use case, I'd recommend using MD5. It's a good balance of speed and collision resistance for cache busting purposes. Here's why:

1. Speed: MD5 is fast to compute, which is important when dealing with potentially large media files.
2. Collision resistance: While not cryptographically secure, it's sufficient for cache busting.
3. Widespread support: MD5 is widely supported and understood.
4. Compatibility: Laravel and PHP have built-in functions for MD5.

caendesilva: Did some benchmarks on the Claude suggestions, and wow CRC32 is fast!

```php
$this->file('zero', '');
$this->file('tiny', 'Hello, World!');
$this->file('small', str_repeat('a', 1024)); // 1 KB
$this->file('compressed', str_repeat('a', 1024 * 256)); // 256 KB
$this->file('medium', str_repeat('a', 1024 * 1024)); // 1 MB
$this->file('large', str_repeat('a', 1024 * 1024 * 10)); // 10 MB

$algos = ['md5', 'sha1', 'sha256', 'crc32'];

// Warm up the statcache
$files = ['zero', 'tiny', 'small', 'compressed', 'medium', 'large'];
array_map('md5_file', array_map(fn ($file) => Hyde::path($file), $files));

$its = 100;

foreach ($algos as $algo) {
    echo "Benchmarking $algo:\n";
    $avgBuffer = [];
    foreach ($files as $file) {
        $result = \Illuminate\Support\Benchmark::measure(function () use ($file, $algo) {
            hash_file($algo, Hyde::path($file));
        }, $its);
        echo "  $file: ".$result."ms\n";
        $avgBuffer[] = $result;
    }
    echo "  Avg: ".array_sum($avgBuffer) / count($avgBuffer)."ms\n";
}
```

```
Benchmarking md5:
  zero: 0.083057ms
  tiny: 0.131709ms
  small: 0.111467ms
  compressed: 0.702308ms
  medium: 2.670385ms
  large: 24.054236ms
  Avg: 4.625527ms
Benchmarking sha1:
  zero: 0.080178ms
  tiny: 0.08563ms
  small: 0.085257ms
  compressed: 0.734365ms
  medium: 2.702944ms
  large: 28.270893ms
  Avg: 5.3265445ms
Benchmarking sha256:
  zero: 0.080278ms
  tiny: 0.085282ms
  small: 0.090875ms
  compressed: 1.401351ms
  medium: 5.457427ms
  large: 54.346367ms
  Avg: 10.243596666667ms
Benchmarking crc32:
  zero: 0.081022ms
  tiny: 0.088007ms
  small: 0.083692ms
  compressed: 0.171689ms
  medium: 0.440915ms
  large: 5.784513ms
  Avg: 1.1083063333333ms
```

Result without the large file:

```
Benchmarking md5:
  zero: 0.08615ms
  tiny: 0.087616ms
  small: 0.086381ms
  compressed: 0.803126ms
  medium: 2.554298ms
  Avg: 0.7235142ms
Benchmarking sha1:
  zero: 0.087621ms
  tiny: 0.091061ms
  small: 0.093293ms
  compressed: 0.788253ms
  medium: 2.759281ms
  Avg: 0.7639018ms
Benchmarking sha256:
  zero: 0.079744ms
  tiny: 0.089832ms
  small: 0.091324ms
  compressed: 1.426645ms
  medium: 5.374404ms
  Avg: 1.4123898ms
Benchmarking crc32:
  zero: 0.082501ms
  tiny: 0.087204ms
  small: 0.084978ms
  compressed: 0.175249ms
  medium: 0.43878ms
  Avg: 0.1737424ms
```

caendesilva: Oh wow, CRC is short! And it's designed for data transmission integrity checks.

```
# hash($algo, 'foo');
md5: acbd18db4cc2f85cedef654fccc4a4d8
sha1: 0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33
sha256: 2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae
crc32: a5c4fe49
```

Now I'm just curious about the hash collision rate, to see if it's a problem in the real world.

caendesilva: > Now I'm just curious about the hash collision rate, to see if it's a problem in the real world.

A site with a large amount of media assets may have 100 files. Given that only a few of those will be updated that much (some images will never be updated, some once or twice, and some things like stylesheets a bunch of times) we can average 3 updates, bringing our estimate to 300 file versions, and 300 hashes.

This brings us to about `1 / 100 000` chance of a hash collision for CRC. That's 1/5th of the chance of being struck by lightning. (If we used SHA1 we'd need 570 trillion times more files to even have a 1 in 100 trillion chance of a collision, which is quite close to the odds of a meteor landing on your house (the odds of which is essentially (because I'm rounding here) the same odds as if we use MD5 on 300 files))

So yeah, considering the worst that can happen is that a cache is not updated (thus also considering in factors of (A user visiting the page containing the asset; A user having visited that page before; The user noticing that an asset is the wrong version) this is not going to be a problem.

image
Source: https://stackoverflow.com/a/14210379/5700388

caendesilva: Note that the "birthday paradox" can come into play here - with just 77,163 files, you have a 50% chance of at least one collision. That's such an edge case, and for any site using that I suggest they change the hash algo if they need it.

Full Claude reference:

Using CRC32 for your static site generator's media file hashing is an interesting choice, especially given its speed advantage when dealing with numerous files. Let's analyze the collision risk and its implications for your use case:

1.  Collision risk with CRC32: CRC32 produces a 32-bit hash, which means there are 2^32 (about 4.3 billion) possible hash values. This is significantly fewer than MD5 (128-bit) or SHA-256 (256-bit), increasing the theoretical chance of collisions.
2.  Practical implications for an SSG:
    -   For cache busting: Collisions are unlikely to cause serious issues. A collision would mean two different files get the same cache-busting key, potentially leading to a cached version being used instead of a new one. This is generally just a temporary inconvenience that resolves itself when the browser fetches the file again.
    -   For file identification: If you're using the hash to uniquely identify files (e.g., in a build process), collisions could potentially cause more significant issues, like overwriting or misidentifying files.
3.  Probability of collisions: The probability of a collision depends on the number of files you're hashing. For a typical website with thousands of media files, the risk is low but not negligible. The "birthday paradox" comes into play here - with just 77,163 files, you have a 50% chance of at least one collision.
4.  Mitigation strategies:
    -   Combine CRC32 with file path: `hash('crc32b', $file_path . filesize($file_path))`
    -   Use a longer CRC variant: CRC64 provides more collision resistance while still being faster than cryptographic hashes.
    -   Implement a collision detection system in your build process.

---

Why combining CRC32 with file path can decrease collisions:

When we combine the CRC32 hash with the file path, we're essentially creating a more unique input for the hash function. Here's why this reduces collisions:
a) Increased input uniqueness: Even if two files have the same content (and thus the same CRC32 hash of their content), they're unlikely to have the same file path.
b) Larger effective hash space: By incorporating the file path, we're increasing the amount of data that contributes to the final hash. This effectively increases the number of possible unique outputs, reducing collision probability.
c) Context-sensitive hashing: The hash now depends not just on the file's content, but also on its location in the file system, making it more specific to your project's structure.

caendesilva: I was thinking we would need a fallback to MD5, but I checked, and CRC32 is available on all supported PHP versions. https://3v4l.org/3L6EJ

caendesilva: For context: Mix seems to use md5 for versioning

caendesilva: https://github.com/hydephp/develop/commit/dca20a021d646a11cbd4b79e13c697cd85c67b5f

caendesilva: Switched in https://github.com/hydephp/develop/commit/dca20a021d646a11cbd4b79e13c697cd85c67b5f

caendesilva: May consider changing the unixsum function to use crc

caendesilva:
## Precompute File Stats at Construct Time

Thinking if we should precompute the metadata like filesize,  mime types, and file hashes, at construct time. I think in the context of the build loop, it would still save time. Assuming that most media files are used in the build at least once[^1], and some are used multiple times in different views (logos, stylesheets, etc, are often used on every page) so if the stats are cached in memory, the time to compute them is only once per file, regardless of how many times they are used in the build.

This would also allow us to do some basic validation of the files at construct time, and fail early if there are issues with the files. The added memory usage of a few dozen bytes per file is negligible. (Note that all of this is essentially micro-optimization, the main benefit is less complexity, but I am making these calculations to ensure there is not a significant downside to the change.)

### Big O Analysis: File Stats Computation

Some calculations on the notation before/after (assuming 25 pages that have 3 assets included on each page (unique ones are static cost and can be disregarded))

Quick napkin math: (See below for breakdown)

```
before: cost = 25 * 3 (pages * assets) = 75 file stats computations
after: cost = 1 * 3 (construct * assets) = 3 file stats computations
```

#### Before

In the before scenario, file stats are computed for each asset on each page:

```
Number of pages: n
Number of assets per page: m
Total computations: n * m

Result: O(n * m) # If we assume m is constant (always 3 assets per page), it simplifies to O(n).
```

#### After

In the after scenario, file stats are computed once for each unique asset at construct time:

```
Number of unique assets: k
Total computations: k

Result: O(k) # Where k is the number of unique assets, which is independent of the number of pages.
```

#### Comparison

-   Before: O(n * m) or O(n) if m is constant
-   After: O(k)

The after scenario has better time complexity, especially as the number of pages (n) grows, assuming the number of unique assets (k) grows more slowly than the number of pages.


[^1]: Fair assumption, people usually add files because they want to use them. Old files that people forgot to clean up happens, but can be ignored from this napkin calculation.

packages/framework/src/Support/Filesystem/MediaFile.php
```diff
@@ -27,6 +29,17 @@ class MediaFile extends ProjectFile
     /** @var array<string> The default extensions for media types */
     final public const EXTENSIONS = ['png', 'svg', 'jpg', 'jpeg', 'gif', 'ico', 'css', 'js'];
 
+    public function __construct(string $path)
+    {
+        $path = trim_slashes(Str::after(Hyde::pathToRelative($path), Hyde::getMediaDirectory()));
+
+        if (Filesystem::missing($path)) {
+            throw new FileNotFoundException($path);
+        }
```
caendesilva, line 38: This should not be possible for this to happen with normal usage, since we generate these based on discovered files. If a developer does something custom, we can't assume they want files backed by disk. We'll see how we handle construct time data generation though...

It also makes unit testing harder (we could add an internal static property to disable validation, or mock filesystem)

caendesilva: We could add validation to the static constructor. But that discrepancy adds more complexity than static property. Filesystem mocks add test complexity

caendesilva:
May want to improve the `HydePage` documentation to match the now better `MediaFile` documentation.

```php
/**
 * Get a collection of all pages, parsed into page models, keyed by the source file identifiers.
 *
 * @return \Hyde\Foundation\Kernel\PageCollection<static>
 */
public static function all(): PageCollection
{
    return Facades\Pages::getPages(static::class);
}
```

PR #1919, opened 2024-07-28
Update the Hyde facade to use a mixin annotation instead of method annotations

caendesilva:
Let's see if this has gotten better support since last time we tried this. If this works, it means that clicking a usage will actually lead to the implementation, and we can get better support for PHPDoc annotations like usages and deprecations.


caendesilva:
image


Gonna do some manual testing on VS Code as well. Currently works great in PHPStorm.

PR #1926, opened 2024-07-31
Update monorepo dependencies

caendesilva:
I think these needs to be merged together:

- Supersedes https://github.com/hydephp/develop/pull/1902
- Supersedes https://github.com/hydephp/develop/pull/1906
- Supersedes https://github.com/hydephp/develop/pull/1925

Additionally it removes some packages from the monorepo that we haven't used in a long time, and are generally better to install through an isolated setup like a dedicated Composer root or using a Phar:

- Remove Rector from the monorepo toolset
- Remove PHP Parallel Lint from the monorepo toolset
- Remove PHP CodeSniffer from the monorepo toolset


composer.json
```diff
@@ -22,15 +22,15 @@
...
     },
     "require-dev": {
-        "vimeo/psalm": "dev-type-coverage as 5.20.0",
         "hyde/monorepo-dev-tools": "*",
         "hyde/realtime-compiler": "*",
         "hyde/testing": "*",
         "jetbrains/phpstorm-attributes": "^1.0",
         "laravel/tinker": "^2.7",
-        "pestphp/pest-plugin-type-coverage": "^2.8",
+        "pestphp/pest-plugin-type-coverage": "2.x-dev",
```
caendesilva, line 30: We need https://github.com/pestphp/pest-plugin-type-coverage/commit/147f8b6d5dc180a756aa694d34f606050b823c71 which has not been released yet. Otherwise we get this:

```
  Too few arguments to function PHPStan\Analyser\NodeScopeResolver::__construct(), 26 passed in /home/runner/work/develop/develop/vendor/pestphp/pest-plugin-type-coverage/src/PHPStanAnalyser.php on line 80 and exactly 27 expected
```

caendesilva:
Not sure why we lost so much type coverage...

caendesilva:
Too many incompatibilities, so gave up with the new Psalm version and reverted to the original.

PR #1932, opened 2024-08-04
[2.x] Improve the media file instance API

caendesilva:
## Abstract 

Makes large improvements to the Asset API https://github.com/hydephp/develop/pull/1904 by creating a more fluent interface for interacting with media assets. It also removes the media link helper (replacing https://github.com/hydephp/develop/pull/1914), now using the Asset API for all media link retrievals. This has the side effect of media files now being required to exist in the project, and that cache busting keys are now consistently added to all media links, and media links will now be qualified using the base URL if set.

### Auxiliary Changes

The filesystem facade has been updated to contain the mime type finder.


caendesilva:
I started adding back support for media files not backed by a file on disk **c48a731551c803150c9f84a05d291b40da6bf8d4,** but realized that won't make sense, because how would they be used? If they have no contents we have nothing to copy to the output directory.

packages/framework/src/Foundation/Kernel/Hyperlinks.php
```diff
@@ -100,28 +102,30 @@ public function mediaLink(string $destination, bool $validate = false): string
             throw new FileNotFoundException($sourcePath);
         }
 
-        return $this->withCacheBusting($this->relativeLink("{$this->kernel->getMediaOutputDirectory()}/$destination"), $destination);
+        return static::withCacheBusting($this->relativeLink("{$this->kernel->getMediaOutputDirectory()}/$destination"), $destination);
     }
 
     /**
      * Gets a relative web link to the given file stored in the `_site/media` folder.
      * If the image is already qualified (starts with `http`) it will be returned as is.
```
caendesilva, line 110: This feature doesn't really make sense, why would we pass a value just to have it be returned the same? Are there (valid) usages where it is not known if the input is for a local or remote file?

caendesilva: Removing this may cause this:

```
Hyde\Framework\Exceptions\FileNotFoundException: File [_media/https://example.com/foo] not found when trying to resolve a media asset.
```

If we keep this we could add an upgrade helper to have a more helpful message if this is detected

caendesilva: Removed in be2e212318cc52a12f83ad9e2ee5c63da85a1e4d

caendesilva:
Had an idea that could be a cool API, if an asset does not exist we return a MissingAsset instance or similar so we can do things like:

```php
function asset(): MediaFile|MissingMediaFile;

if (asset('css/app.css')->exists()) {
    // Do something 
}
```

caendesilva:
It's kinda a discrepancy between these that you can link to anything regardless of if it exists or not, but assets throw. But I think it's fair balancing the DX and complexity. Just wanted to note it.

```php
$this->assertSame('foo', Hyde::relativeLink('foo'));
$this->assertSame('media/foo', (string) Hyde::asset('foo'));
```

PR #1933, opened 2024-08-05
[2.x] Lazy loaded MediaFile metadata properties

caendesilva:
## Abstract

- Follow up to https://github.com/hydephp/develop/pull/1918 
- This is a part of https://github.com/hydephp/develop/pull/1904

## Changes

- Makes the public readonly metadata properties protected and only accessible via getter methods
- Lazy loads metadata properties using a boot system, which constructs all metadata properties when one is accessed (and when instance is serialized to array), and then caches them for future access
- No longer validates file existence in the constructor, but instead on boot, making it possible to create memory based file objects which opens up the API for more use cases, at the cost of inconsistent places where the file existence is checked but still having a robust system in place for when trying to get filesystem information for a non-existent file

### DIFF
--- a/packages/framework/src/Support/Filesystem/MediaFile.php
+++ b/packages/framework/src/Support/Filesystem/MediaFile.php
@@ -32,8 +32,6 @@ class MediaFile extends ProjectFile
     public function __construct(string $path)
     {
         parent::__construct($this->getNormalizedPath($path));
-
-        $this->boot();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/framework/src/Support/Filesystem/MediaFile.php
```diff
@@ -97,27 +97,21 @@ public function toArray(): array
 
     public function getContentLength(): int
     {
-        if (! isset($this->length)) {
-            $this->boot();
-        }
+        $this->ensureInstanceIsBooted('length');
```
caendesilva, line 100: Technically we could do something like `$this->getBootedProperty()` but I think this is much more readable and understandable.

PR #2006, opened 2024-11-10
[2.x] New asset system

caendesilva:
## Abstract

This pull request for HydePHP v2 brings several improvements on how we handle assets in HydePHP. It brings a complete rewrite of the HydeFront asset handling system, replacing Laravel Mix with Vite, and favouring Blade-based components with Tailwind classes over CSS partials and custom stylesheets.

### Introduces Vite

In these pull requests we replaced Laravel Mix with Vite:

- [x] https://github.com/hydephp/develop/pull/2010
- [x] https://github.com/hydephp/develop/pull/2011
- [x] https://github.com/hydephp/develop/pull/2012
- [x] https://github.com/hydephp/develop/pull/2013
- [x] https://github.com/hydephp/develop/pull/2016
- [x] https://github.com/hydephp/develop/pull/2021
- [x] https://github.com/hydephp/develop/pull/2060

### HydeFront v4

We are updating how HydeFront is handled for v2. Instead of declaring styles directly in HydeFront alongside Tailwind, we will refactor those styles into Tailwind. 

HydeFront will serve two main purposes:  

1. It will continue to include the precompiled `app.css` file bundled with new apps, available through the HydeFront CDN.  
2. It will act as a component library, allowing users to include granular styles from `app.css`, which we will preconfigure.  

For example, users can include our Tailwind styles granularly using just the `app.css` file from the HydeFront package. This file will be compiled alongside Tailwind. If users prefer customization, they can remove the import and add the specific styles they want.  

- [x] https://github.com/hydephp/develop/pull/2024 
  - [x] https://github.com/hydephp/develop/pull/2029
  - [x] https://github.com/hydephp/develop/pull/2031
  - [x] https://github.com/hydephp/develop/pull/2036
  - [x] https://github.com/hydephp/develop/pull/2037
  - [x] https://github.com/hydephp/develop/pull/2038
  - [x] https://github.com/hydephp/develop/pull/2039

### Related Changes

- [x] https://github.com/hydephp/develop/pull/2056
- [x] https://github.com/hydephp/develop/pull/2045
- [x] https://github.com/hydephp/develop/pull/2047

PR #2010, opened 2024-11-11
[2.x] Replace Laravel Mix with Vite

caendesilva:
## Abstract

- Fixes https://github.com/hydephp/develop/issues/1129
- Supersedes https://github.com/hydephp/develop/pull/1130

**Important:** You must now use `npm run build` instead of `npm run prod`

### Motivation

Key changes from Laravel Mix to Vite:

- Simpler configuration with Vite
- Faster development builds with HMR support
- Better production optimization
- Modern ESM-based development
- Match new Laravel syntax of `npm run build`.
- And we can use HMR as a BrowserSync replacement.

### Community support

Vite is also winning by a landslide by a [community vote](https://x.com/CodeWithCaen/status/1855587933241086318)

[image
](https://x.com/CodeWithCaen/status/1855587933241086318)

## Laravel Mix vs Vite Benchmark

All cases have had the NPM command been run before the benchmark to warmup any caches. Note that the benchmarks did not run enough iterations to get an accurate average. Tests were run on Windows.

### Laravel Mix:

- **First build:** `9694 ms`
- **Rebuild:** `8907 ms`
- **Average:** `9300 ms`

### Vite:

- **First build:** `4628 ms`
- **Rebuild:** `4518 ms`
- **Average:** `4575 ms`

### Analysis

Vite is approximately 52% faster than Laravel Mix, in other words it's about twice as fast.


caendesilva:
Since `run dev` has a changed scope, we may want to remove this: (a merged `--vite` option may work better)

```php
if ($this->option('run-dev')) {
    $this->runNodeCommand('npm run dev', 'Building frontend assets for development!');
}
```

And update this:

> Then run `npm run dev` to compile the assets in development mode. For production builds, run `npm run build`.

caendesilva:
## Clause response to renamed build flag

Based on the provided code and context, I'll analyze the situation and provide some suggestions for renaming the flags.

Current situation:

```php
    protected $signature = 'build
        {--run-dev : Run the NPM dev script after build}
        {--run-prod : Run the NPM prod script after build}
        {--run-prettier : Format the output using NPM Prettier}
        {--pretty-urls : Should links in output use pretty URLs?}
        {--no-api : Disable API calls, for example, Torchlight}';
```


The current flags `--run-dev` and `--run-prod` are misleading now that we've switched to Vite, since:

1. `npm run dev` with Vite starts a development server
2. `npm run build` with Vite builds for production
3. There's no direct equivalent to Laravel Mix's unminified build

Here are my suggestions, ordered by preference:

### 1. `--compile-assets`
Motivation:
- Generic and future-proof name that describes what it does
- Doesn't tie to specific build tool terminology
- Single flag simplifies the API
- Clear that it's about asset compilation
- Matches Hyde's existing terminology (e.g. "compile static pages")

### 2. `--build-assets`
Motivation:
- Aligns with Vite's `npm run build` command
- Simple and clear purpose
- Common terminology in frontend tooling
- Matches documentation that talks about "building assets"

### 3. `--vite-build`
Motivation:
- Explicitly mentions the tool being used
- Clear that it's using Vite's build command
- Matches the actual npm script name
- But: Could be limiting if we switch tools again

I recommend going with `--compile-assets` because:

1. It's the most tool-agnostic option
2. Simplifies the API by consolidating two flags into one
3. Matches Hyde's existing terminology
4. Clearly communicates purpose to users
5. Won't need changing if build tools change again

The implementation would then just run `npm run build` since we no longer need separate dev/prod modes with Vite.

caendesilva:
Think I will go with `--run-vite` to match `--run-prettier`. I think it's clear enough that we mean it runs the Vite compiler/bundler, not the server, given the context of a final site build.

PR #2016, opened 2024-11-12
[2.x] Add Vite Realtime Compiler integration for HMR

caendesilva:
## Add Vite Realtime Compiler integration for HMR

### Description

This PR integrates the Vite with the Realtime Compiler for Hot Module Replacement for a highly efficient development workflow by enabling real-time updates without needing a full page reload.

### Key Changes

- **Added Vite facade:** A new Vite facade is added for handling Vite-related operations, allowing for easy access to the Vite server status and for generating Vite asset tags.
- **Vite integration in serve command:** The `serve` command now has a `--vite` option to enable Vite for Hot Module Replacement (HMR). When enabled, the command starts a Vite development server along with the HydeRC server.
- **Asynchronous process handling:** The `serve` command now runs the server and Vite processes asynchronously, allowing for parallel execution and more responsive output.
- **Output streaming:** The console output from both the server and Vite processes is streamed directly to the terminal.
- **Port availability check:** The `serve` command also checks if the Vite server port is available before starting the Vite development server, preventing conflicts and providing useful error messages.

This significantly improves the developer experience by allowing for fast, iterative development with instant feedback via HMR.

### Future Tasks

- [ ] Better output https://github.com/hydephp/develop/pull/2016#issuecomment-2476362113
- [ ] Fix Windows performance https://github.com/hydephp/develop/pull/2016#issuecomment-2476444190
- [ ] Normalize include location https://github.com/hydephp/develop/pull/2016#discussion_r1842413128


caendesilva:
Can probably add nicer output, but maybe in a later PR.

image

caendesilva:
Resources for live reload later on https://github.com/andrefelipe/vite-php-setup/blob/master/vite/vite.config.js

caendesilva:
> We have this latter part to automatically connect to a Vite server directly run through `npm run dev` instead of `php hyde serve --vite`.
> 
> ```
> // Check if Vite dev server is running by attempting to connect to it
> // Todo: Improve performance on Windows (takes less than 1ms on macOS, but around 100ms on Windows)
> set_error_handler(fn () => false); // Todo: This warning surpressor does not work on Windows
> $server = fsockopen('localhost', 5173, $errno, $errstr, 0.1);
> restore_error_handler();
> 
> if ($server) {
>     fclose($server);
> 
>     return true;
> }
> 
> return false;
> ```
> 
> However, this introduces an unacceptable real-time compiler performance penalty for Windows users, where each page load now takes 100ms longer as this check is performed every time a page is served.
> 
> Here are my ideas initial ideas to adress this.
> 
> * While we can't cache this as a static property since each previewed page is run in a new process, we could, from the realtime compiler serve command, do this check there and add the env var if running. But this means that Vite would only work if that command was run before the serve command, unless we use the output loop to periodically check for a Vite server, but that would then require the web server to restart in order to get the new env var.
> * Add a "hotfile" like Laravel uses?
> * Vite manifest trickery?

## Claude Analysis 

Let me analyze the options and propose a solution:

### Analysis of Options

1. **Environment Variable from Serve Command**
- Pros: Simple to implement
- Cons: Requires server restart, not reliable for direct Vite usage

2. **Hot File Approach** (Laravel-style)
- Pros: 
  - Fast (file_exists check is much quicker than fsockopen)
  - Works across processes
  - Can be updated in real-time
- Cons: 
  - Needs cleanup handling

3. **Vite Manifest**
- Pros: Already part of Vite workflow
- Cons: Not suitable as it's only created after build

4. **New Option: Combined Approach**
- Use hot file created by serve command
- Periodic check in output loop for direct Vite usage
- File-based check instead of socket connection

### Recommended Solution: Hot File Approach

This is the best option because:
1. Eliminates the performance penalty of socket connections
2. Works across processes
3. Supports both serve command and direct Vite usage
4. Similar to proven Laravel approach

### Development Plan

1. **Create ViteManager Class**
```php
namespace Hyde\Framework\Services;

class ViteManager
{
    protected static function getHotFile(): string
    {
        return Hyde::path('storage/framework/cache/vite.hot');
    }

    public static function createHotFile(): void
    {
        file_put_contents(static::getHotFile(), time());
    }

    public static function removeHotFile(): void
    {
        @unlink(static::getHotFile());
    }

    public static function isRunning(): bool
    {
        return file_exists(static::getHotFile());
    }
}
```

2. **Update Vite Facade**
```php
public static function running(): bool
{
    // Check if Vite was enabled via the serve command
    if (env('HYDE_SERVER_VITE') === 'enabled') {
        return true;
    }

    return ViteManager::isRunning();
}
```

3. **Update ServeCommand**
Reference:

```46:85:packages/framework/src/Console/Commands/ServeCommand.php
    protected ConsoleOutput $console;

    protected InvokedProcess $server;
    protected InvokedProcess $vite;

    public function safeHandle(): int
    {
        $this->configureOutput();
        $this->printStartMessage();

        if ($this->option('open') !== 'false') {
            $this->openInBrowser((string) $this->option('open'));
        }

        $command = sprintf('php -S %s:%d %s',
            $this->getHostSelection(),
            $this->getPortSelection(),
            $this->getExecutablePath()
        );

        if ($this->option('vite')) {
            $this->runViteProcess();
        }

        $this->runServerProcess($command);

        while ($this->server->running()) {
            if (isset($this->vite) && $this->vite->running()) {
                $output = $this->vite->latestOutput();

                if ($output) {
                    $this->output->write($output);
                }
            }

            Sleep::usleep(100000); // 100ms
        }

        return Command::SUCCESS;
    }
```


Add hot file management:
```php
protected function runViteProcess(): void
{
    ViteManager::createHotFile();
    
    // Existing Vite process code...
    
    register_shutdown_function(function () {
        ViteManager::removeHotFile();
    });
}
```

4. **Add Periodic Check in Output Loop**
```php
while ($this->server->running()) {
    if (isset($this->vite) && $this->vite->running()) {
        ViteManager::createHotFile();
    } else {
        ViteManager::removeHotFile();
    }
    
    // Existing output handling...
}
```

### Implementation Benefits

1. Windows performance improves significantly as `file_exists()` is much faster than `fsockopen()`
2. Supports both `php hyde serve --vite` and direct `npm run dev` usage
3. Auto-cleanup through shutdown function
4. No need for server restarts
5. Maintains current functionality while improving performance

packages/framework/resources/views/layouts/styles.blade.php
```diff
@@ -2,17 +2,22 @@
 <style>[x-cloak] {display: none!important}</style>
 
 {{-- The compiled Tailwind/App styles --}}
-@if(config('hyde.load_app_styles_from_cdn', false))
-    <link rel="stylesheet" href="{{ HydeFront::cdnLink('app.css') }}">
-@elseif(Asset::exists('app.css'))
-    <link rel="stylesheet" href="{{ Asset::get('app.css') }}">
-@endif
+@if(Vite::running())
+    {{ Vite::assets(['resources/assets/app.css']) }}
```
caendesilva, line 6: Also needs the script

packages/framework/src/Console/Commands/ServeCommand.php
```diff
@@ -169,4 +194,28 @@ protected function getOpenCommand(string $osFamily): ?string
...
+    {
+        $socket = @fsockopen('localhost', $port, $errno, $errstr, 1);
+        if ($socket !== false) {
+            fclose($socket);
+
+            return false;
+        }
+
+        return true;
+    }
```
caendesilva, line 220: Let's mark this feature as `@experimental`

packages/realtime-compiler/resources/dashboard.blade.php
```diff
@@ -98,7 +98,7 @@
                                                     await navigator.clipboard.writeText(data);
                                                     document.getElementById("copyPathToClipboardButtonIcon").style.display = 'none';
                                                     document.getElementById("copyPathToClipboardButtonIconSuccess").style.display = 'inline';
-                                                    await new Promise(resolve => setTimeout(resolve, 3000)).then(function () {
+                                                    await new Promise(resolve => setTimeout(resolve, 5173)).then(function () {
```
caendesilva, line 101: Whoops.

vite.config.js
```diff
@@ -6,8 +6,35 @@ import { defineConfig } from 'vite';
...
+                            path.resolve(__dirname, 'vendor/hyde/realtime-compiler/resources/vite-index-page.html'),
+                            'utf-8'
+                        ));
+                    } else {
+                        next();
+                    }
+                });
+            },
+        },
+    ],
```
caendesilva, line 37: Eventually we may want to extract a NPM plugin for this

packages/framework/resources/views/layouts/styles.blade.php
```diff
@@ -2,17 +2,22 @@
...
+@else
+    @if(config('hyde.load_app_styles_from_cdn', false))
+        <link rel="stylesheet" href="{{ HydeFront::cdnLink('app.css') }}">
+    @elseif(Asset::exists('app.css'))
+        <link rel="stylesheet" href="{{ Asset::get('app.css') }}">
+    @endif
+
```
caendesilva, line 13: Would it make more sense if we just automatically enabled Vite when using `Asset::get` and Vite is running?

packages/framework/resources/views/layouts/scripts.blade.php
```diff
@@ -1,5 +1,7 @@
 {{-- The compiled Vite scripts --}}
-@if(Asset::exists('app.js'))
+@if(Vite::running())
```
caendesilva, line 2: May make more sense to have this alongside the CSS include since the script is deferred anyways, but not sure how to handle that best with the partials since it would fit better in the `head` component, and don't want to make such a big change here.

PR #2024, opened 2024-11-15
[2.x] Refactor HydeFront styles to Tailwind

caendesilva:
### Abstract

We are updating how HydeFront is handled for v2. Instead of declaring styles directly in HydeFront alongside Tailwind, we will refactor those styles into Tailwind. 

HydeFront will serve two main purposes:  

1. It will continue to include the precompiled `app.css` file bundled with new apps, available through the HydeFront CDN.  
2. It will act as a component library, allowing users to include granular styles from `app.css`, which we will preconfigure.  

For example, users can include our Tailwind styles granularly using just the `app.css` file from the HydeFront package. This file will be compiled alongside Tailwind. If users prefer customization, they can remove the import and add the specific styles they want.  

Todo: Test that checks if config files are matching should test the Tailwind config

### Related Changes

- [x] https://github.com/hydephp/develop/pull/2029
- [x] https://github.com/hydephp/develop/pull/2031
- [x] https://github.com/hydephp/develop/pull/2036
- [x] https://github.com/hydephp/develop/pull/2037
- [x] https://github.com/hydephp/develop/pull/2038

I think changelog in many parts can be summarized as:
- Refactored several Blade components to use Tailwind instead of custom HydeFront CSS code


caendesilva:
Can tweak this a bit more
Old https://images.percy.io/f212d2f4b690bb850b3f8c529644b8c0e61dedc7d2f55ef01ae000421e507419
New https://images.percy.io/b403a7fff64a77c15b7614442e07e69782040d29ea4c79a5bbae9f7166af5fb3

caendesilva:
There are going to be some tiny visual regressions/changes, so I'm going to tackle all of those later. Want to get the main refactor done first.

resources/assets/app.css
```diff
@@ -17,4 +17,26 @@
...
+    @apply block -ml-8 pl-8 opacity-80 hover:opacity-100 hover:bg-gray-200/20 transition-all duration-300 relative;
+}
+
+.table-of-contents a::before {
+    @apply content-['#'] text-[75%] opacity-50 mr-1 transition-opacity duration-300;
+}
+
+.table-of-contents a:hover::before {
+    @apply opacity-100;
+}
```
caendesilva, line 42: I think later I want to refactor the table of contents generation to be Blade-based. The generator can get the data needed, and a customizable Blade partial is used to add the styles.

resources/assets/app.css
```diff
@@ -17,4 +17,51 @@
...
+    @apply opacity-0 ml-1 transition-opacity duration-300 ease-linear px-1 scroll-m-4;
+}
+
+.heading-permalink::before {
+    @apply content-['#'];
+}
+
+.heading-permalink:hover, .heading-permalink:focus {
+    @apply opacity-100 grayscale-0;
+}
```
caendesilva, line 67: Also, ideally if we could have a post-processor (or custom CommonMark heading renderer?) to swap out headings to a customizable Blade component to fully support changing all parts of heading permalinks from Blade would be great, for example moving them around and stuff would be really nice. Also supporting changing the icons and stuff too of course.

resources/assets/app.css
caendesilva: Now tracked in https://github.com/hydephp/develop/pull/2006

resources/assets/app.css
caendesilva: Now tracked in https://github.com/hydephp/develop/pull/2006

resources/assets/app.css
caendesilva: Fixed in https://github.com/hydephp/develop/pull/2045

resources/assets/app.css
caendesilva: Now handled in https://github.com/hydephp/develop/pull/2047

PR #2029, opened 2024-11-16
[2.x] Recreate the HydeSearch plugin with Alpine.js

caendesilva:
Replaces the [HydeSearch](https://github.com/hydephp/HydeSearch) plugin with an Alpine.js and Blade implementation to make it more customizable and to reduce external dependencies. It also lets us more easily couple fixes and improvements to Hyde versions.

- Targets https://github.com/hydephp/develop/pull/2006


caendesilva:
I think it makes sense to hide the keyboard shortcut help text when there are search results, for a few reasons:

1. Once users are actively searching/viewing results, the shortcuts become less relevant
2. The space could be better used for showing more results
3. There are multiple obvious ways to exit the modal:
   - The ESC key (which is a common pattern across the web)
   - Clicking the X button in the header
   - Clicking outside the modal (click-away behavior)
   - The search button is still visible

PR #2045, opened 2024-12-01
[2.x] Blade-based table of contents generator

caendesilva:
This allows the table of contents to be customized by publishing the Blade component, and is now styled using real TailwindCSS.

It's also up to 20 times faster.


caendesilva:
Undefined array key "slug" with this:

```md
## test
### test
```

PR #2047, opened 2024-12-01
[2.x] Custom Markdown heading renderer

caendesilva:
## Abstract

- Fixes https://github.com/hydephp/develop/pull/2024#discussion_r1844957700

### Changes

- Removed auto-configuration for `HeadingPermalinkExtension`, meaning user added custom options provided to that extension will be ignored. If you for some reason require that extension's custom features, either implement those to the Blade component after publishing it, or disable the Hyde permalink feature from the configuration and add the extension yourself.
- Removes the `MarkdownService` `withPermalinks` and `canEnablePermalinks` methods, [see comment](https://github.com/hydephp/develop/pull/2047#issuecomment-2510243423)
- https://github.com/hydephp/develop/pull/2052


caendesilva:
#### Cursor (Claude) analysis of the three approaches for implementing heading permalinks:

### 1. Original CommonMark Extension Implementation

```207:221:packages/framework/src/Framework/Services/MarkdownService.php
    protected function configurePermalinksExtension(): void
    {
        $this->addExtension(HeadingPermalinkExtension::class);
                'fragment_prefix' => '',
        $this->config = array_merge([
            'heading_permalink' => [
                'id_prefix' => '',
                'fragment_prefix' => '',
                'symbol' => '',
                'insert' => 'after',
                'min_heading_level' => 2,
                'aria_hidden' => false,
            ],
        ], $this->config);
    }
```


Benefits:
- Simple to implement and maintain
- Follows CommonMark's standard extension pattern
- Performant since it's handled during Markdown parsing
- Consistent behavior across all Markdown content

Drawbacks:
- Limited customization options
- Requires modifying CommonMark configuration
- Can't easily change permalink position/style per heading
- Tightly coupled to CommonMark implementation

### 2. Full Blade Component Replacement

```1:13:resources/views/components/heading-permalink.blade.php
@props(['level' => 1, 'id' => null])

@php
$tag = 'h' . $level;
$id = $id ?? \Illuminate\Support\Str::slug($slot);
@endphp

<{{ $tag }} {{ $attributes->merge(['id' => $id]) }}>
    {{ $slot }}
    @if(config('markdown.features.permalinks', true))
        <a href="#{{ $id }}" class="heading-permalink" aria-label="Permalink for this section"></a>
    @endif
</{{ $tag }}> 
```


Benefits:
- Full control over heading HTML structure
- Can customize behavior per heading instance
- Easy to modify styling and structure
- Can add additional features like custom icons
- More maintainable Blade templating

Drawbacks:
- More complex implementation
- Requires replacing CommonMark's heading renderer
- Slightly higher performance overhead
- Need to maintain heading level logic
- Must handle ID generation manually

### 3. Permalink-Only Component Injection

Example implementation:
```php
// components/heading-permalink.blade.php
@props(['id'])

<a href="#{{ $id }}" 
   class="heading-permalink" 
   aria-label="Permalink for this section">
</a>
```

Benefits:
- Minimal implementation
- Easy to inject before/after existing headings
- Keeps CommonMark's heading rendering intact
- Very focused responsibility
- Lightweight performance impact

Drawbacks:
- Less control over heading structure
- Need to coordinate with existing heading HTML
- May require additional DOM manipulation
- Could have positioning challenges
- Limited contextual awareness

### Recommendation

I recommend using the full Blade component replacement approach (#2) because:

1. It provides the most flexibility for future customization
2. Keeps the implementation in pure Blade/PHP without CommonMark dependencies
3. Allows per-heading customization when needed
4. Easier to maintain and modify
5. Better integration with Laravel/Blade ecosystem
6. Can still use existing CSS without changes:

```1:20:packages/hydefront/components/heading-permalinks.css
.prose h1, .prose h2, .prose h3, .prose h4, .prose h5, .prose h6 {
    @apply w-fit;
}

.prose :is(h1,h2,h3,h4,h5,h6):hover .heading-permalink,
.prose :is(h1,h2,h3,h4,h5,h6):focus .heading-permalink {
    @apply opacity-75 grayscale transition-opacity duration-100 ease-out;
}

.heading-permalink {
    @apply opacity-0 ml-1 transition-opacity duration-300 ease-linear px-1 scroll-m-4;
}

.heading-permalink::before {
    @apply content-['#'];
}

.heading-permalink:hover, .heading-permalink:focus {
    @apply opacity-100 grayscale-0;
}
```


The slight performance overhead is negligible for most sites, and the benefits of having full control over the heading structure outweigh the drawbacks. It also makes it easier to add features like custom icons, positioning options, and accessibility improvements in the future.

The permalink-only component approach might seem simpler at first, but it introduces complexity in positioning and coordination with existing heading elements. The full component replacement gives us a clean slate to work with while maintaining all the functionality we need.

packages/framework/resources/views/components/markdown-heading.blade.php
```diff
@@ -0,0 +1,13 @@
+@props(['level' => 1, 'id' => null, 'addPermalink' => config('markdown.features.permalinks', true)])
+
+@php
+    $tag = 'h' . $level;
+    $id = $id ?? \Illuminate\Support\Str::slug($slot);
+@endphp
+
+<{{ $tag }} {{ $attributes->merge(['id' => $id]) }}>
+    {!! $slot !!}
```
caendesilva, line 9: Confirmed this does not open up for XSS.

Example input:

```md
# Quickstart Guide <script>console.log('Quickstart Guide');</script> <strong>bold</strong> <em>italic</em>
```

```html
<h1 id="quickstart-guide-ltscriptconsolelogquickstart-guideltscript-strongboldstrong-emitalicem">
    Quickstart Guide &lt;script&gt;console.log('Quickstart Guide');&lt;/script&gt; <strong>bold</strong> <em>italic</em>
            <a href="#quickstart-guide-ltscriptconsolelogquickstart-guideltscript-strongboldstrong-emitalicem" class="heading-permalink" aria-label="Permalink for this section"></a>
    </h1>
```

image

caendesilva:
This will also remove the `MarkdownService::withPermalinks` method, since neither I nor Claude Sonnet via Cursor can justify its existence.

---

Based on the provided code and context, I recommend removing the `withPermalinks()` method for the following reasons:

1. **No Active Usage**: Looking through the codebase, this method is only referenced in tests and not used in any actual application code or documented features.

2. **Redundant Functionality**: The permalink functionality is already handled through other means:
- Configuration via `markdown.permalinks.enabled` setting
- The `canEnablePermalinks()` method which has proper logic for determining if permalinks should be enabled
- Documentation pages automatically get permalinks when table of contents is enabled

3. **Maintenance Burden**: Keeping this method would require:
- Maintaining compatibility with the new permalinks system
- Documenting the feature
- Testing edge cases and interactions with other permalink-related features
- Potentially confusing users by providing multiple ways to enable the same functionality

4. **Clean Architecture**: Removing unused methods helps keep the codebase cleaner and reduces cognitive load for developers maintaining the system.

5. **Breaking Change Opportunity**: Since this is for v2, it's an ideal time to remove unused APIs that could become technical debt.

The relevant test that would need to be removed or modified is in:

```217:225:packages/framework/tests/Feature/Services/HydeSmartDocsTest.php
    public function testClassCanParseDocumentWithDisabledPermalinks()
    {
        config(['markdown.permalinks.enabled' => false]);

        $article = $this->makeArticle("# Header Content \n\n Body Content");

        $this->assertSame('<h1>Header Content</h1>', $article->renderHeader()->toHtml());
        $this->assertSame('<p>Body Content</p>', $article->renderBody()->toHtml());
    }
```


I recommend removing the `withPermalinks()` method and updating any related tests to use the proper configuration methods instead. This will simplify the codebase without removing any actual functionality that users depend on.

caendesilva:
**Not sure which is best:**

Which method of heading permalinks is best for a markdown renderer?

```html
        <h2 id="documentation-heading">Documentation Heading <a href="#documentation-heading" class="heading-permalink" title="Permalink">#</a></h2>
        <h2>Documentation Heading<a id="documentation-heading" href="#documentation-heading" class="heading-permalink" title="Permalink">#</a></h2>
```

---

**GPT-4o answer:**

Both methods can work for generating heading permalinks in a Markdown renderer, but each has its advantages and trade-offs depending on the goals of your renderer. Here's a breakdown of the two approaches:

---

### **1st Method: Using `id` in the `<h2>` tag**
```html
<h2 id="documentation-heading">Documentation Heading <a href="#documentation-heading" class="heading-permalink" title="Permalink">#</a></h2>
```

#### **Advantages:**
- **Cleaner semantics:** The `id` directly ties to the heading itself, which aligns well with HTML5 standards.
- **Improved accessibility:** Assistive technologies can better associate the `id` with the heading content.
- **CSS styling:** Easier to target headings via their `id` for custom styles.

#### **Disadvantages:**
- **Duplicated text:** The permalink anchor (`<a>` with `#`) is part of the heading text, potentially causing confusion if copy-pasting.
- **Requires extra spacing:** Styling is needed to visually separate the `#` permalink from the heading text.

---

### **2nd Method: Using `id` in the `<a>` tag**
```html
<h2>Documentation Heading<a id="documentation-heading" href="#documentation-heading" class="heading-permalink" title="Permalink">#</a></h2>
```

#### **Advantages:**
- **Copy-paste clarity:** The heading text can be copied without including the permalink character.
- **Simplified DOM structure:** Keeps the heading content separate from metadata like the `id`.

#### **Disadvantages:**
- **Less accessible:** The `id` is associated with the `<a>` rather than the heading, which might be less intuitive for assistive technologies.
- **Potential styling complexity:** Aligning the anchor visually and semantically with the heading may require extra CSS.

---

### **Recommendation:**
The **1st method** is generally better for most Markdown renderers due to its semantic correctness and accessibility. While the permalink being part of the text might be a minor annoyance for some users, this can be mitigated with good CSS (e.g., hiding the permalink on copy or adjusting its appearance).

However, if your primary goal is a cleaner user experience for copy-pasting (e.g., documentation sites where users frequently copy headings), the **2nd method** might be preferable.

packages/framework/resources/views/components/markdown-heading.blade.php
caendesilva: And tested in https://github.com/hydephp/develop/commit/43d82a4327dcf2cd68d02a97cec041bdcd8dbe28

caendesilva:
Edge case we need to fix: If there are multiple headings of the same name, we need offset.

caendesilva:
Was wondering if we should have a config option for the hash symbol. Until users request it I think I'm going to hold off on that. Claude's (via Cursor) explanation below is very in line with my thoughts.

---

Based on the context and the nature of this customization, I recommend requiring users to publish the blade component rather than adding a configuration option. Here's why:

1. The hash symbol (#) is a widely recognized standard for permalink/anchor links across the web
2. Looking at the configuration files, particularly `config/markdown.php`, we can see the philosophy is to keep configuration options focused on essential functionality rather than visual customizations
3. The existing documentation (as seen in `docs/digging-deeper/customization.md`) already guides users on how to publish and customize Blade views when needed
4. This aligns with Laravel's approach where visual/UI customizations are typically handled through view publishing rather than configuration

If someone wants to customize the permalink icon, they can:

1. Run `php hyde publish:views` 
2. Edit the published component in `resources/views/vendor/hyde/components/markdown-heading.blade.php`
3. Change the "#" to their desired icon/text

This approach:
- Keeps the configuration clean and focused on core functionality
- Provides more flexibility since users can add HTML/CSS classes along with changing the icon
- Follows the established pattern for UI customizations in Hyde/Laravel
- Avoids adding rarely-used configuration options

The documentation already shows this is the recommended approach for UI customizations, as seen in:


````77:90:docs/creating-content/managing-assets.md
### Customizing the Blade templates

To make it really easy to customize asset loading, the styles and scripts are loaded in dedicated Blade components.

- Styles are loaded in `hyde::layouts.styles`
- Scripts are loaded in `hyde::layouts.scripts`

To customize them, run the following command:

```bash
php hyde publish:views layouts
```

Then edit the files found in `resources/views/vendor/hyde/layouts` directory of your project.
````

caendesilva:
Just for absolute fun, not that I thought this would in any way be a performance sink, but the code that generates unique identifiers takes only `0.011ms` for a heading registry with `1 000 000` items.

PR #2052, opened 2024-12-04
[2.x] Semantic Tailwind Markdown heading permalinks

caendesilva:
- Part of https://github.com/hydephp/develop/pull/2047
- Also refactors to use Tailwind instead of CSS

**Advantages:**

-   Cleaner semantics: The `id` directly ties to the heading itself, which aligns well with HTML5 standards.
-   Improved accessibility: Assistive technologies can better associate the `id` with the heading content.
-   CSS styling: Easier to target headings via their `id` for custom styles.

See https://github.com/hydephp/develop/pull/2047#issuecomment-2511210214


caendesilva:
Edge case we need to fix: If there are multiple headings of the same name, we need offset.

PR #2060, opened 2024-12-08
[2.x] Improve the Vite integration

caendesilva:
Targets https://github.com/hydephp/develop/pull/2006

- Refactors code
- Adds documentation
- Adds more tests
- Throws if the path is not a supported extension (Laravel assumes non-CSS is JavaScript... I'd rather be explicit)
- Removes the internal `HYDE_SERVER_VITE` environment variable
  - The complexity this adds to microoptimize things is not warranted yet
  - The Vite server configuration already manages adding and removing the hotfile.
  - <s>**We need this to print the start message. Instead, it renames the variable to be more clear it's internal**</s> Edit: Fixed by just touching the hotfile right away

### DIFF
--- a/packages/framework/src/Facades/Vite.php
+++ b/packages/framework/src/Facades/Vite.php
@@ -22,11 +22,11 @@ class Vite
 
         foreach ($paths as $path) {
             if (str_ends_with($path, '.css')) {
-                $html .= sprintf('<link rel="stylesheet" href="http://localhost:5173/%s">', $path);
+                $html .= static::formatStylesheetLink($path);
             }
 
             if (str_ends_with($path, '.js')) {
-                $html .= sprintf('<script src="http://localhost:5173/%s" type="module"></script>', $path);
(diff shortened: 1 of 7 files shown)


packages/framework/src/Facades/Vite.php
```diff
@@ -5,37 +5,70 @@
...
 use Illuminate\Support\HtmlString;
+use InvalidArgumentException;
 
 /**
  * Vite facade for handling Vite-related operations.
  */
 class Vite
 {
+    protected const CSS_EXTENSIONS = ['css', 'less', 'sass', 'scss', 'styl', 'stylus', 'pcss', 'postcss'];
+    protected const JS_EXTENSIONS = ['js', 'jsx', 'ts', 'tsx'];
```
caendesilva, line 16: This is seriously overkill, but it doesn't really add any more complexity to add all these. While they might not work unless more media file extensions are registered, adding all these here future proofs things without needing to support customization here.

PR #2062, opened 2024-12-10
[2.x] Interactive component publisher command

caendesilva:


packages/framework/src/Console/Helpers/InteractivePublishCommandHelper.php
```diff
@@ -0,0 +1,143 @@
...
+            Filesystem::copy($source, $target); // Todo: See how we should handle existing files
+        }
+    }
+
+    /**
+     * @experimental This method may be toned down in the future.
+     *
+     * @param  array<string>  $selectedFiles
+     */
+    public function formatOutput(array $selectedFiles): string
```
caendesilva, line 101: Is this too far?


All files (10): "Published all files, including [app.blade.php, docs.blade.php, footer.blade.php] and 7 more"
Many files (7): "Published files [app.blade.php, docs.blade.php, footer.blade.php] and 4 more"
Multiple files (3): "Published files [app.blade.php, docs.blade.php, footer.blade.php]"
Single file (1): "Published file [app.blade.php]"

caendesilva:
Ahh, looks like we don't need the mock wrapper as Laravel Prompts does indeed support testing (but says nothing on it on the website). It's just Laravel Zero being tricky. https://github.com/laravel/prompts/issues/39#issuecomment-1667053448

caendesilva:
Figuring out how to do Laravel Prompts properly, so will automatically enable this when available in the terminal.

caendesilva:
In order to reduce complexity, and open up for another feature, I think this command will only support groups of files. Publishing the 404 page will be made in a new `template:publish` command, which also supports homepages and others.

caendesilva:
Passed manual testing on macOS and Windows.

Issue #2063, opened 2024-12-11
Replace glob brace usages with more robust solutions


caendesilva:
Bumping the priority of this. We need to replace glob brace usages as some Linux distros do not support it, for instance Alpine, which hurts Docker support.

Related issues:

- https://github.com/hydephp/framework/issues/662
- https://github.com/hydephp/hyde/issues/270

PR #2064, opened 2024-12-11
[1.x] Replace `GLOB_BRACE` with a more robust solution

caendesilva:
Replaces glob brace usages as some Linux distros do not support it, for instance Alpine, which hurts Docker support.

Related issues: (Which this will probably fix)

- Presumed to fix https://github.com/hydephp/framework/issues/662
- Confirmed to fix https://github.com/hydephp/hyde/issues/270

Will fix https://github.com/hydephp/develop/issues/2063


caendesilva:
I think it makes sense for a file finder to handle file extensions in a case-insensitive manner, especially in scenarios where the target operating system or use case does not enforce case sensitivity. This is particularly important because:

1.  **Cross-Platform Consistency**: Different operating systems handle case sensitivity differently. For example:

    -   Windows and macOS treat file extensions as case-insensitive.
    -   Linux is case-sensitive. Making the file finder case-insensitive ensures predictable behavior across all platforms.
2.  **User Convenience**: Users might unintentionally create files with mixed-case extensions (e.g., `.MD` vs `.md`). A case-insensitive approach avoids confusion and ensures files are found regardless of how the extension is written.

3.  **Practicality**: When searching by extension, the intent is generally to match all files of a certain type, not to differentiate based on casing.

caendesilva:
I think we should sort the output in the method too to further normalize different file systems. It takes just `0.010ms` to sort 100 files, so that's fully acceptable.

caendesilva:
Better to have paths relative to the project root to follow the conventions, and how we will use it. Especially since it's a drop in replacement for the smart glob method which is relative to the project.

caendesilva:
Okay, I think this is ready for merge, and assuming the root cause of the two suspected related issues is `GLOB_BRACE`, this should fix those.

- @clifmo could you confirm if this branch fixes your issue https://github.com/hydephp/framework/issues/662? (unless the cause was Docker being weird, this should fix your issue, as it doesn't match directories)
- @sobi3ch could you confirm if this branch fixes your issue https://github.com/hydephp/hyde/issues/270 (pretty sure it will)


To test this, run the following command in your project terminal:

```bash
composer require hyde/framework:dev-replace-glob-brace-preview
```

Then rerun the reproduction steps that led to your issues.

If either of you don't have time to test it, that's totally okay, just let me know. I will be going out of town for a few days and will hold of on merging this until then as I can't properly support a release rollout until the end of the week.

caendesilva:
> @caendesilva works like a charm 

@sobi3ch Thank you so much for taking the time to test it! I'll be going ahead and merge this, and I will try for a weekend release, if not earlier!

PR #2067, opened 2024-12-21
Internal: Retain commit authors for split monorepo packages

caendesilva:
**Key changes made:**

1.  Added new environment variables for author name and email from the head commit
2.  Updated Git config to use the original author's name and email instead of github-actions
3.  Experimental: Makes it so that it says that the original user authored and GitHub Actions committed

This will now attribute the split repository commits to the original author of the commit being synced.

---

**Background**

This workflow currently makes the split commit as the GitHub Actions user. This was originally so that if there are multiple commits, we may not want to attribute unrelated commits to authors. But I had an idea:

First, I was thinking that if the commit is a merge commit (like from a pull request), we can use the commit author then. Because if an author merges a PR, they are presumed to be responsible for all commits in that PR, so we can safely attribute the merge to them (as in essence, this split is like a squash merge).

The same if it's just a single commit we are syncing, we can keep the user.

But then, after looking at the commit history, I realized we only ever really have single changes because the merge queue makes so only one change goes through at a time, so the split monorepo job will just run on the last pushed/merged code. So, we can use the author of the last commit.

So, this updates all jobs in the workflow to use the original commit author for the latest commit (the one we're basing the message on).

### DIFF
--- a/.github/workflows/split-monorepo.yml
+++ b/.github/workflows/split-monorepo.yml
@@ -65,6 +65,8 @@ jobs:
       - name: Commit and push changes
         env:
           COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
+          COMMIT_AUTHOR_NAME: ${{ github.event.head_commit.author.name }}
+          COMMIT_AUTHOR_EMAIL: ${{ github.event.head_commit.author.email }}
         run: |
           cd hyde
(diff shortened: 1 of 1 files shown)


caendesilva:
Yup, looks nice.

image

PR #2141, opened 2025-04-20
[2.x] Upgrade to Laravel 11

caendesilva:


caendesilva:
Okay for some reason we're losing type coverage collection. With an hour of debugging I can't fix it, and I can't spend more time on that. Just noting it here that Psalm stopped playing nice. Our(my) fault for using a fork. Will fix later.

PR #2163, opened 2025-04-27
Internal: Fix GitHub action that switched to ESM

caendesilva:
Previous versions of actions/github-script (v6 and earlier) used CommonJS internally, where require() worked fine. Now it uses ESM, so the action fails.


caendesilva:
Didn't work. Trying again.

PR #2165, opened 2025-04-27
Internal: Inline buggy workflow scripts

caendesilva:
Since this didn't help https://github.com/hydephp/develop/pull/2164


caendesilva:
Yup, this worked!

Issue #1743, opened 2024-06-28
Simplify Routes facade methods


caendesilva:
> Wondering if we should take more steps to normalize the API, and match to Laravel. I'm not sure if there is ever really a reason to want a null return here, I think if you don't want to catch an exception, it's better to use Routes::exists to check first. See https://github.com/hydephp/develop/pull/591 which was closed due to similar motivations

_Originally posted by @caendesilva in https://github.com/hydephp/develop/issues/1731#issuecomment-2196324574_

caendesilva:
The current methods as follows:

```php
public static function get(string $routeKey): ?Route
{
    return static::getFacadeRoot()->get($routeKey);
}

/** @throws \Hyde\Framework\Exceptions\RouteNotFoundException */
public static function getOrFail(string $routeKey): Route
{
    return static::getFacadeRoot()->getRoute($routeKey);
}
```

Could be merged and simplified like this:

```php
public static function get(string $routeKey): Route
{
    return static::getFacadeRoot()->getRoute($routeKey);
}
```

Considering this is quite a BC break for quite a small thing, we would need to carefully consider this

caendesilva:
I think it's worth considering this change for v2, since it's our only window to make BC breaks for a while. Simplifying the Routes facade to follow Laravel conventions makes the API more predictable—get() would always throw on failure, and users can use exists() if they want a safe check. This removes redundancy (get vs getOrFail) and encourages clearer code. While it's a small change, it improves consistency, and the impact can be mitigated with a clear upgrade note.

caendesilva:
Fixed in https://github.com/hydephp/develop/pull/2179

PR #2176, opened 2025-05-10
[2.x] Merge route helper methods

caendesilva:
# PR Description:
This PR simplifies the Routes facade API by merging the `get()` and `getOrFail()` methods into a single `get()` method that always throws a `RouteNotFoundException` when a route doesn't exist. This aligns with Laravel's conventions and makes the API more predictable. While we lose the ability to use null coalesces, the benefit in predictability and uniformity outweighs that syntax sugar.

## Changes:

- Merged `get` and `getOrFail` into a single `get` method that throws an exception on failure
- Updated code that relied on the null return behavior to use `Routes::exists()` checks first
- Updated tests and documentation to reflect the new behavior

## Benefits:
- More predictable API - `get()` will always throw if the route doesn't exist
- Better alignment with Laravel's conventions
- Encourages explicit existence checks using `Routes::exists()` rather than null coalescing
- Removes redundancy in the API

## Breaking Changes:
- Code that relied on `Routes::get()` returning null for non-existent routes needs to be updated to use `Routes::exists()` first
- A complete upgrade guide is provided in the documentation

This implements the proposal from #1743.


caendesilva:
I think a better option is to add something like tryGet() method that would return null for non-existent routes for developers who prefer that pattern. So basically a rename

caendesilva:
Superseded by https://github.com/hydephp/develop/pull/2179

Issue #1240, opened 2023-03-08
Remove support for dot notation in route key normalization


caendesilva:
Currently periods are rewritten to slashes when accessing route keys, this is so that dot notation can be used, which provides familiarity to those coming from Laravel where that's the convention for route names.

However, this leads to severe issues when there are actually periods in the URL. For example: `http://localhost:8080/docs/1.x/index.html` would be parsed as `http://localhost:8080/docs/1/x/index.html` which results in `Route [docs/1/x/index] not found.`

This then begs the question if we should support normalizing route keys to support dot notation at all, especially since I think it's a fairly common use case to use HydePHP for documentation sites, which are often versioned like this. Furthermore, [periods are allowed in URLs](https://stackoverflow.com/questions/7555553/can-period-be-part-of-the-path-part-of-an-url).

It also adds complexity to have multiple route key namings, and since route keys are directly tied to the URL routes I think it makes more sense to only use that format.

See https://github.com/hydephp/develop/pull/1239

caendesilva:
This is basically a side effect of not having manually named routes.

PR #2177, opened 2025-05-10
Fix lack of dot notation support caused by faulty test

caendesilva:


caendesilva:
Remembered that we removed this in https://github.com/hydephp/develop/issues/1240, so the test just needs to be removed.

PR #2235, opened 2025-05-29
feat: forward method calls to DateTime in DateString

Cyberking99:
This PR adds support for forwarding method calls from the `DateString` class to its internal `DateTime` instance. This allows for more intuitive usage, such as calling `$post->date->format('Y-m-d')` directly instead of accessing `$post->date->dateTimeObject->format('Y-m-d')`.

- [ ] Implemented the __call() magic method to intercept and forward undefined method calls to the DateTime object.
- [ ] Throws a BadMethodCallException if a method does not exist on DateTime, maintaining robustness.

### DIFF
--- a/packages/framework/src/Support/Models/DateString.php
+++ b/packages/framework/src/Support/Models/DateString.php
@@ -46,4 +46,13 @@ class DateString implements Stringable
     {
         return $this->short;
     }
+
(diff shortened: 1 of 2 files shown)


caendesilva:
Looks great! Make sure to unmark your PR as draft and fix style issues when you're ready for this to be merged 🔥

PR #2239, opened 2025-05-30
Fix rebuild command showing the wrong console output

nats-afs:
output message about file location fixed

Fixes https://github.com/hydephp/hyde/issues/288


caendesilva:
@nats-afs If you want, feel free to make a second PR with the same change to the master branch so we get this in the next HydePHP v2 release candidate as well 🔥

nats-afs:
Nice
This is the [second PR](https://github.com/hydephp/develop/pull/2240)

PR #2418, opened 2026-04-13
Fix realtime compiler loading assets from production URL

### LINKED ISSUE #2327: [Realtime Compiler] Images trying to load from production URL when on localhost
For example featured image was not loading locally, request was made to https://hydephp.com/media/v2-cover-min.png?v=401c1839. I think we could set the URL to be the request URL (localhost or Herd) to improve this. Should not be any side effects since the output is no longer saved.

emmadesilva:
When the realtime compiler served pages with a configured site URL (e.g. hydephp.com), asset and media file references were generated as absolute URLs pointing to the production host, causing the browser to request files from production instead of the local server.

Override the configured site URL to the current request's host:port when booting the application for page compilation. This ensures all generated asset links resolve against localhost (or the configured dev server) rather than production.

Skip the override when save_preview is enabled to avoid baking the local URL into persisted output files.

Closes #2327


emmadesilva:
Confirmed locally this works well!

Issue #2456, opened 2026-06-27
[v3] Add command-level config overrides for build commands


emmadesilva:
## Summary

Add a generic way to override Hyde config values from the CLI when running build-related commands.

Instead of adding more dedicated command options for every possible setting, v3 could support repeatable config override arguments, for example:

```bash
hyde build --config=hyde.pretty_urls=true
hyde build --config=hyde.features.play_cdn=true
hyde rebuild --config=hyde.pretty_urls=true --config=hyde.features.play_cdn=false
```

This would make it possible to temporarily change build behavior for a single command without editing the project config files.

## Motivation

Some config values affect site builds, but may also be useful to change for a single command invocation.

For example, there is a config option for using the Tailwind Play CDN instead of compiling Tailwind assets. That setting affects site builds, not only the realtime compiler. Currently, adding dedicated CLI flags for every such option could lead to a large number of command arguments.

A generic config override option would scale better than adding many one-off flags.

## Proposed API

Add a repeatable `--config` option to build-related commands:

```bash
hyde build --config=key=value
hyde rebuild --config=key=value
```

Multiple overrides should be supported:

```bash
hyde build \
  --config=hyde.pretty_urls=true \
  --config=hyde.features.play_cdn=true
```

Suggested precedence:

```text
CLI config override > command-specific option > project config > default value
```

In practice, command-specific options such as `--pretty-urls` could internally use the same temporary config overlay system.

## Example use cases

Build once with pretty URLs enabled:

```bash
hyde build --config=hyde.pretty_urls=true
```

Build once using the Tailwind Play CDN:

```bash
hyde build --config=hyde.features.play_cdn=true
```

Rebuild with multiple temporary config values:

```bash
hyde rebuild \
  --config=hyde.pretty_urls=true \
  --config=hyde.features.play_cdn=false
```

## Implementation tips

Laravel Artisan supports repeatable option values using option arrays.

The command signature could use:

```php
protected $signature = 'build
    {--config=* : Override config values for this command, e.g. --config=hyde.pretty_urls=true}';
```

Then each provided value can be parsed as `key=value`:

```php
foreach ($this->option('config') as $override) {
    if (! str_contains($override, '=')) {
        $this->error("Invalid config override [$override]. Expected key=value.");

        return self::FAILURE;
    }

    [$key, $value] = explode('=', $override, 2);

    config()->set($key, $this->parseConfigValue($value));
}
```

A simple value parser could support booleans, null, numbers, and strings:

```php
[...]

PR #2489, opened 2026-07-03
[3.x] Add build command config flags

emmadesilva:
Fixes https://github.com/hydephp/develop/issues/2456

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This serves two purposes:
 
 ### Added
 - Added a configuration option to disable the footer scroll-to-top button independently of the footer in https://github.com/hydephp/develop/pull/2459
+- Added a repeatable `--config=key=value` option to the `build` and `rebuild` commands, letting you temporarily override a config value for a single command invocation without editing the project config files. The overrides take precedence over both the project config and other build options, such as `--pretty-urls`.
 
 ### Changed
 - for changes in existing functionality.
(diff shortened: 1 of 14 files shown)


emmadesilva:
Closing this — after thinking it through against Hyde's "simplicity first, power when you need it" motto, a generic `--config=key=value` cuts against both halves for `build`/`rebuild`, and isn't needed for `serve`.

**Why:**

- **Reproducibility.** `build`'s whole job is to faithfully materialize what's on disk. A generic override means the same commit can produce a different site depending on an invisible flag nobody wrote down — the exact "which environment produced this?" ambiguity we don't want. Config-on-disk should stay the single source of truth for anything build-affecting.
- **Type safety / quality.** The `key=value` parser is lossy by construction — no way to distinguish a string `"true"` from the boolean `true`, or set a value that's literally the string `"123"`. That's not a quality bar we want on a mechanism that mutates arbitrary config keys, some of which (per the issue) may not even be safe to touch at runtime.
- **We already have the power lever.** `HYDE_PRETTY_URLS` / `HYDE_PLAY_CDN` env vars plus `serve`'s curated flags (`--pretty-urls`, `--play-cdn`, `--dashboard`) already cover the actual use case (dev-time convenience without touching prod config). A generic override on top would just be a fourth competing mechanism, less disciplined than the three we already have.
- **build vs. serve aren't the same kind of command.** Normalizing their CLI surface was the original motivation, but `build` produces the deployed artifact and `serve` is throwaway/dev-only — treating them as siblings papers over that difference rather than respecting it.

**Net:** a curated set of named flags (like `serve` already has) beats a generic escape hatch. More flags is a better trade than ambiguity about what actually built the site.

Not pursuing this for v3. If a genuinely uncovered `serve`-time setting comes up, we'll add it as its own named flag, consistent with `--pretty-urls`/`--play-cdn`. The `rebuild`-accepts-`build`-flags parity fix is still worth doing separately as a patch.

PR #2500, opened 2026-07-04
[3.x] New publish command

emmadesilva:
Collapses the four-command publishing surface into a single `publish` command, and moves raw config publishing onto a standard Laravel tag.

```bash
php hyde publish                  # interactive: views, or a starter page
php hyde publish --layouts        # scope the picker to layouts
php hyde publish --components     # scope the picker to components
php hyde publish --all            # every view, no picker

php hyde publish --page           # starter page picker
php hyde publish --page=welcome   # publish a named starter page
php hyde publish --page=posts --to=_pages/index.blade.php

php hyde vendor:publish --tag=hyde-config
```

## Views

Views are published to `resources/views/vendor/hyde`. One grouped multi-select picker offers layouts and components together, with an "All views" row; `--layouts` and `--components` prefilter it, and `--all` skips it. Output is cardinality-aware, reporting what was copied, what was already current, and what was left alone.

## Starter pages

Starter pages are backed by a new `PublishablePages` registry, so a page can carry a label, a default destination, alternate destinations, and its own Hyde page class. The catalog ships `welcome`, `posts`, `blank`, and `404`.

There is no separate homepage concept any more — setting a homepage is publishing a homepage-capable page to `_pages/index.blade.php`. Destinations resolve from `--to`, the page's default, or an interactive prompt when the choice is genuinely ambiguous. A `--to` path is validated against the selected page's own class, so `source_root`, custom source directories, file extensions, and extended page classes are all respected rather than assuming `_pages/*.blade.php`. Two pages resolving to the same destination are caught before anything is written.

Packages can register their own pages through `PublishablePages::register()`.

## Safe by default

Publishing no longer overwrites your work. A shared `OverwritePolicy` decides per file: missing → copy, identical → skip, modified → blocked. Comparison is EOL-agnostic, so a CRLF checkout does not read as a modification. Blocked files need either an interactive confirmation or `--force`. Every outcome is decided before the first write, so cancelling never leaves a half-published tree.

This is the main behavior change from v2, where all three publish commands overwrote silently.

## Config

Config publishing was rare and has fixed destinations, so it becomes a plain tag rather than a bespoke command:

```bash
php hyde vendor:publish --tag=hyde-config
```

That publishes `hyde.php`, `docs.php`, `markdown.php`, `view.php`, `cache.php`, and `commands.php`. Torchlight is excluded — it comes from Torchlight's own package tag. The legacy `configs`, `hyde-configs`, and `support-configs` tags are removed.

## Breaking changes

`publish:views`, `publish:homepage`, and `publish:configs` are removed outright rather than aliased. Invoking one raises Symfony's native command-not-found error, which already suggests `publish`. The full replacement table — including the `posts` and `blank` mappings that need `--to=_pages/index.blade.php` to keep publishing to the homepage — is in the upgrade guide.

Removing them also retires `AsksToRebuildSite`, `ViewDiffService` (and its checksum index), `InteractivePublishCommandHelper`, and the per-homepage `vendor:publish` tags. The `hyde-page-404` tag is kept: unlike the homepage tags, it had no removed consumer, so dropping it would be a separate user-facing change.

`publish` deliberately exposes no provider, tag, or path publishing. `--tag`, `--provider`, and `--config` are intercepted before Symfony's option binding and redirected to `vendor:publish`, without being advertised in the command's help output. A genuine typo like `--layout` still surfaces Symfony's own error.

## Notes for review

- Commits are atomic and meant to be read in order: policy, registry, helpers, views flow, pages flow, command, config tag, removals, docs, tests.
- `_pages/index.blade.php` and `packages/framework/resources/views/homepages/welcome.blade.php` are byte-identical, and the page tests depend on that. Both were updated together where the welcome copy mentions the old command.
- A `--to` path containing a `..` segment is rejected rather than normalized away.
- The CI gh-pages step and the Dusk browser tests were updated; both previously called `publish:homepage`.


emmadesilva:
Closing due to severely diverged branch and am recreating it

PR #2502, opened 2026-07-07
[3.x] Introduce Blade Blocks for Markdown pages

emmadesilva:
## Abstract

This pull requests adds a processor for all Markdown pages that allows you to put blocks of Blade and call Blade components from Markdown pages.

This takes a simplified approach compared to the [Bladedown](https://github.com/hydephp/bladedown) proof of concept, where this is more predictable and blockbased instead of purely blending the languages. The latter has some interesting syntax, but this PR gives an easier mental model.

## Details

<html><head></head><body><p>This pull request adds a Markdown processor that lets trusted Markdown pages contain blocks of Blade and call Blade components directly from Markdown.</p><p>This is a sister feature to Hyde's <code inline="">[Blade]:</code> BladeDown support, but takes a simpler and more predictable block-based approach. Instead of blending the languages together, this keeps Markdown as Markdown and only executes Blade when a fenced code block explicitly opts into it.</p><p>The feature is opt-out through <code inline="">markdown.enable_blade_blocks</code> in <code inline="">config/markdown.php</code>. Like BladeDown, this can execute arbitrary PHP, so it should only be enabled for trusted Markdown content.</p><p>Blade is evaluated at build time, and every executed block receives the <code inline="">$page</code> variable so the block can access the current page's data. Rendered Blade blocks are wrapped in:</p><pre><code class="language-html">&lt;div class="blade-block not-prose"&gt;
</code></pre><p>A bare <code inline="">blade</code> block is intentionally not executed. It remains a normal highlighted code sample:</p><pre><code class="language-markdown">```blade
{{ "This is not executed" }}
```
</code></pre><p>Anything executable is keyed off the <code inline="">blade</code> prefix plus a directive. This means we still get Blade syntax highlighting, while avoiding surprising behavior for ordinary Blade examples.</p><p>For arbitrary inline Blade, you can use <code inline="">blade render</code>:</p><pre><code class="language-markdown">```blade render
@php($world = 'world')

{{ "Hello $world" }}
```
</code></pre><p>This renders through <code inline="">Blade::render()</code> and works well for small anonymous snippets.</p><p>You can also call Blade components from Markdown using a <code inline="">blade component(name)</code> directive with YAML front matter-style props:</p><pre><code class="language-markdown">```blade component(component-name)
foo: bar
```
</code></pre><p>For components with body content, the front matter can be enclosed with triple dashes. The remaining content is compiled as Markdown and passed as the component slot:</p><pre><code class="language-markdown">```blade component(component-name)
---
foo: bar
---

# I can now write Markdown here!
```
</code></pre><p>The slot Markdown is compiled using Hyde's Markdown converter, so normal Markdown features still apply there, including things like enhanced code blocks. Since the rendered block itself is wrapped in <code inline="">not-prose</code>, component authors may still want to add <code inline="">prose</code> to their own slot containers when appropriate.</p><p>Normal code blocks are unaffected, and Markdown inside component slots does not render Blade expressions, so this stays scoped to the outer Blade block:</p><pre><code class="language-markdown">```blade component(component-name)
---
foo: bar
---

Also, in component Markdown slots, the {{ $blade }} is not rendered.
```
</code></pre><p>If you need to write documentation that contains an actual Blade block example, you can use four backticks around it so the inner block is shown instead of processed.</p><p>A few edge cases are handled explicitly:</p>
Block | Behavior
-- | --
blade | highlighted sample
blade render | valid Blade block
blade component(name) | valid Blade block
blade component | throws an error
blade component() | throws an error
blade component(foo | throws an error
blade foo | throws an error
php, js, html | ordinary code block

<p>Blocks with the same content are compiled independently, so repeated render blocks are evaluated separately rather than being reused or cached by content identity.</p></body></html>


emmadesilva:
I'm actually thinking now that this is actually such a clean syntax, we could ship this in all Markdown pages, and we already have a feature called "Blade-supported Markdown" ('markdown.enable_blade') in the config the Blade directive that we could probably retire after this since I can NEVER remember the syntax...

PR #2575, opened 2026-07-31
[3.x] Support setting titles in terminal code blocks

emmadesilva:
Adds an optional title to terminal code blocks, so the window's title bar can describe what the block is showing instead of always reading "Terminal".

````markdown
```terminal title="Installing Hyde"
$ composer require hyde/framework
```
````

Combined with the existing `xml` modifier:

````markdown
```terminal xml title="Build output"
<info>Hyde was installed successfully.</info>
```
````

The info string grammar is now:

```
terminal [xml] [title="…"]
```

Modifiers are order-independent, so `terminal xml title="Build output"` and `terminal title="Build output" xml` are the same thing.

## Why

`title="…"` is the established convention for this exact thing. Docusaurus uses it for titled fenced code blocks, and Expressive Code uses it specifically for terminal window titles, so readers coming from other documentation generators recognize it without being taught. It also fits CommonMark, which treats the first word of the info string as the language and deliberately leaves the rest for implementations to interpret.

Within Hyde it is the same attribute syntax as the Blade Block component directive (#2574), which is what the extensibility argument there anticipated: the title sits next to the existing modifier rather than replacing or reshaping it.

## Quoting rules

Following the component directive: double quotes canonical, single quotes equivalent (useful when the title itself contains a double quote), and unquoted or unterminated values rejected rather than guessed at. Unlike a component name, a title may contain whitespace, which is exactly why it is quoted; the quotes keep the rest of the info string parseable as space-separated tokens.

| Input | Result |
|---|---|
| `terminal title="Build output"` | Canonical |
| `terminal title='Build output'` | Accepted |
| `terminal title='The "build" command'` | Accepted |
| `terminal title=""` | Accepted, renders an empty title bar |
| `terminal title=Build` | Throws |
| `terminal title="Build output` | Throws |

Rejecting a malformed title departs from how the block treats an unknown modifier, which is still ignored so that a modifier added in a future version does not break a page rendered by an older one. The asymmetry is deliberate: `title=Build` is not a modifier this version doesn't know about, it is one it does know about, written wrong, and silently discarding it would leave an author looking at a window still labelled "Terminal" with nothing to explain why.

## View contract

The title is passed to the view verbatim as a nullable string rather than pre-resolved to the default label, so a published view can define its own fallback for untitled blocks:

```blade
<span>{{ $title ?? 'Terminal' }}</span>
```

An explicitly empty title is therefore distinguishable from an omitted one. The documented customization example now edits this fallback instead of hardcoding a label, so per-block titles keep working in a customized view.

## Notes

Terminal blocks have not shipped in a release, so this is additive rather than a breaking change. A published view from earlier in the v3 cycle keeps working untouched; it simply ignores the new variable.

Documented on both the Advanced Markdown syntax reference and the composable blocks masterclass, with every new claim asserted in the documentation driven test. The commits are split so each one stands on its own: the feature with its tests, the documentation, then the planning note recording the rationale.

### DIFF
--- a/HYDEPHP_V3_PLANNING.md
+++ b/HYDEPHP_V3_PLANNING.md
@@ -114,7 +114,14 @@ space-separated tokens.
 Rejecting a malformed title departs from how the block treats an unknown modifier, which is ignored so that a modifier
 added in a future version does not break a page rendered by an older one. The asymmetry is deliberate: `title=Build` is
 not a modifier this version doesn't know about, it is one it does know about, written wrong, and silently discarding it
-would leave an author looking at a window still labelled `Terminal` with nothing to explain why.
+would leave an author looking at a window still labelled `Terminal` with nothing to explain why. The same reasoning
+covers a bare `title`, a space around the `=`, and a second title on the same block: each is a title written wrong
+rather than a modifier from the future, so each is reported instead of ignored.
+
+The info string is tokenized by walking a cursor through it, requiring each token to end where the next one begins,
+rather than searching it for modifiers wherever they appear. That is what makes the rejections above reliable:
+`title="One"xml` is one malformed token instead of two valid modifiers, and a modifier is only recognized when it
+stands on its own, exactly as the documented grammar describes it.
 
 The title is passed to the view verbatim as a nullable string rather than pre-resolved to the default label, so that a
 published view can define its own fallback for untitled blocks, which the shipped view demonstrates with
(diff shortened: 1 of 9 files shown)


emmadesilva:
Fixed in e33acb3c3e — the tokenizer was the real issue.

The info string is now tokenized by walking a cursor through it with `\G`, requiring each token to end at a whitespace boundary or end-of-string, instead of searching the string for modifiers wherever they appear. `title="One"xml` is now a single bare token rather than two valid modifiers, and since that token starts with `title=`, it is rejected.

Malformed title detection was widened to match: a bare `title` token throws too, which covers `title = "Build"` and `title ="Build"` (the `title` token is malformed on its own, before the parser ever reaches the rest).

| Input | Before | Now |
|---|---|---|
| `title="Build"xml` | Accepted as two modifiers | Throws |
| `title="Build"future="yes"` | Accepted as two modifiers | Throws |
| `title` | Ignored | Throws |
| `title = "Build"` | Ignored | Throws |
| `title ="Build"` | Ignored | Throws |
| `title="First" title="Second"` | Last one wins | Throws |
| `xmlfuture="yes"` | Enabled `xml` | Ignored as one unknown token |

I took the duplicate-title suggestion as well, for the reason you gave: two different titles are two conflicting intentions, so `A terminal block can only have one title.` is thrown rather than documenting an arbitrary last-wins rule. A repeated flag like `xml xml` is still fine, since a repeated flag has only one possible meaning.

Unknown modifiers are still ignored, both bare (`future`) and as attributes (`future="yes"`), so a modifier added in a future version does not break a page rendered by an older one. That asymmetry is now stated explicitly in the documentation rather than being implied.

Regression tests cover every row above, in both `TerminalCodeBlocksTest` and the documentation driven test. Both documentation pages and the planning note were updated: modifiers are documented as whitespace-separated, one title per block, and rejection now covers "anything that reads as a title but isn't a quoted `title="…"` attribute" rather than just unquoted and unterminated values.

emmadesilva:
Agreed — the tokenizer was disproportionate. Reverted it and took the regex approach instead (e8cc26f494).

The whole change is now two assertions on the existing pattern plus one condition:

```php
protected const TOKEN_PATTERN = '/(?<!\S)(?:(?<key>[\w-]+)=(?:"(?<double>[^"]*)"|\'(?<single>[^\']*)\')|(?<word>\S+))(?=\s|$)/';
```

```php
if ($word === 'title' || str_starts_with($word, 'title=')) {
```

`title="One"xml` now backtracks out of the attribute branch into `\S+`, becoming one bare token that the existing rejection catches. The bare `title` condition covers `title = "Build"` and `title ="Build"`, since the first token is malformed before the parser reaches the rest.

Duplicate titles are left as last-one-wins, undocumented and untested, as you suggested. No `$titled` state, no extra rule.

Documentation is two amended sentences rather than new paragraphs: "unquoted, or that never closes its quote" became "unquoted, never closing its quote, or spaced around the `=`" on both pages. The planning note gained one clause on why tokens are matched between boundaries.

Tests are three more entries in the existing malformed-title list, two in the documentation driven one, and a single new case asserting that `xmlfuture="yes"` is one unknown token rather than an `xml` modifier — which is the behavior the assertions actually buy. The parser is still at 100% patch coverage, without an unreachable branch this time.

Net diff against the version you reviewed first: 6 files, +31/-12.

PR #2576, opened 2026-07-31
[3.x] Render terminal blocks through a view model

emmadesilva:
## Summary

Terminal blocks now render through a view model, `Hyde\Framework\Views\Components\TerminalBlockComponent`, instead of a renderer assembling an array of view data. The contract previously only existed as the arguments of a `view()` call, so the node, the renderer, and the view each had to agree on it while nothing declared it for a reader, an IDE, or the documentation.

It is a Laravel `Illuminate\View\Component` rather than a data object of our own, since that contract already exists and is already understood: public properties become view data, published views keep taking precedence, and attribute bags, subclassing, and `<x-…>` registration come along at no cost.

## What this changes

- **The contract is declared.** `$contents`, `$literal`, `$title`, `$usesSymfonyFormatting`, and `$attributes` are the component's public properties, and they are exactly what the view receives (Blade's own bookkeeping properties are excluded).
- **A terminal window can be rendered from anywhere**, without writing Markdown:
  ```php
  echo (new TerminalBlockComponent($process->getOutput(), title: 'Build output'))->toHtml();
  ```
  `withAttributes()` merges attributes onto the `<figure>`, and a project that wants the tag can register the class in a service provider.
- **The node carries the component** instead of duplicating its properties, which keeps one source of truth for the block's shape and leaves `TerminalBlockRenderer` as a thin adapter between CommonMark and the component. The formatting work moved from the renderer into the component.
- **`$contents` is an `HtmlString`**, so a view can echo it with either syntax without double-escaping it, and `$literal` is passed beside it for views that need the unrendered text (a copy button, a plain text fallback).

## Documentation

- The Composable Markdown Blocks masterclass gains a *View models* section, a rewritten terminal view contract, and a *Rendering terminal windows from PHP* section. The "write your own block" walkthrough now teaches the same pattern the framework uses.
- The Advanced Markdown page points at the new section.
- The release notes describe the terminal block as it will ship, since none of this has been released; the reasoning behind the component is recorded in the planning notes.

## Tests

15 new tests across `TerminalCodeBlocksTest` and the documentation driven `ComposableMarkdownBlocksDocumentationTest`, covering the view data contract, the `HtmlString` semantics, `$literal` escaping, attribute merging, published view resolution, and Blade tag registration. Full suite, HydeStan, and the DocGen linter are green.


emmadesilva:
Thanks — I verified each finding against the code rather than reasoning from the diff. Pushed 8246380 addressing 1, 2, 4, and 5. Details, including the two I'm not taking:

### 1. Leaked methods on the documented callout — confirmed, fixed

Reproduced: the example's view data was `contents,type,componentName,attributes,toHtml,ignoredParameterNames`.

I took your stronger suggestion rather than the `$except` blacklist, for both the framework component and the walkthrough example: `data()` is now declared explicitly, so the contract can't drift as public members are added. The docs gained a short *Declaring the view data* note explaining why, since the walkthrough teaches this pattern, and `$except` is mentioned there as the lighter alternative. New test: `testAddingAPublicMemberDoesNotAddAViewVariable` subclasses the component with a public property and a public method and asserts neither reaches the view.

### 2. `toHtml()` bypasses `shouldRender()` — confirmed, fixed

Verified a `shouldRender(): false` subclass still rendered through the direct path. `toHtml()` now checks it. Covered by `testComponentDoesNotRenderWhenItShouldNotBeRendered`.

### 4. Order-sensitive assertion — resolved

With `data()` declared explicitly the order *is* the contract, so `assertSame` is now the right assertion rather than an incidental one. Kept as-is on that basis.

### 5. Wording — fixed

- "The constructor takes the same data the table above describes" → now says it takes the output and its two modifiers, with an explicit note that `$contents` is derived and `$attributes` is the inherited bag.
- "Every public property … and nothing else is passed" → now describes the declared view data, plus a paragraph on the slot variables the `<x-…>` path adds. Confirmed empirically: tag rendering yields `…,attributes,slot,__laravel_slots`, the direct path stops at `attributes`. Both are now asserted by tests.
- Softened the double-escaping claim.
- The planning-file contract list now includes `$attributes`.

### 3. Component construction in the transformer — keeping as-is

The concrete part of this doesn't reproduce. `Markdown::render()` resolves a **fresh** `MarkdownService` through the container each call, so the nested render in the callout example doesn't re-enter the converter being parsed — there's no shared state to corrupt, and no event-ordering interaction. The remaining formatting work is string manipulation on a block that is about to be rendered anyway, so nothing is computed early or wasted.

That leaves the architectural preference, and moving construction into the renderer costs the thing the PR is for: the node would restate the block's shape, and the renderer would restate it again as constructor arguments — three places to keep in step, which is the duplication the view model removes. I've written the trade-off into the planning notes so the choice is on the record rather than implicit, and per your alternative the node's role is stated there too.

Full suite green at 3666 passed (up 60 assertions), HydeStan and the DocGen linter clean.

emmadesilva:
Reverting most of the previous commit in eebd4b6 — I over-corrected on finding 1.

Overriding `data()` to seal the view data made this an unusual Blade component (normal ones never override it), and the documentation section justifying it taught people to work around Laravel rather than use it. That's backwards for this feature: these blocks are worth building on precisely because what you already know about Blade applies, and the PR's own framing says so.

The real defect was a documentation claim that was too absolute — "only its public properties reach the view". The fix for an over-claim is to soften the claim, not to reshape the code so the claim becomes true. Extra variables in the view data are harmless; every Blade component in every Laravel project has them.

So:

- `data()` override removed from both the framework component and the walkthrough example, and the *Declaring the view data* section is gone.
- The contract table now introduces itself as "the variables the block gives the view", with a line noting Blade's own variables are there too, as for any component.
- Dropped the slot-variables paragraph, which only existed to qualify the exclusivity claim.
- The documentation test now asserts the useful direction — every variable documented in the table is actually given to the view — instead of asserting exclusivity.

**Kept from that commit:** `toHtml()` still checks `shouldRender()`. That one is a genuine behavioural fix rather than a framework workaround, and it makes the direct path agree with the tag path.

Planning notes record the rejected alternative and why, so this doesn't get re-litigated later.

Suite green at 3662, HydeStan and DocGen clean.

PR #2577, opened 2026-08-01
[3.x] Render fenced code blocks through a composable Blade view

emmadesilva:
## Summary

Fenced code blocks were the last Markdown construct in Hyde still assembled by string surgery. A pre-processor rewrote `// filepath:` comments into `<!-- HYDE[Filepath] -->` markers, and a post-processor spliced the rendered label back into the finished HTML with a regex matching `<pre><code class="language-…">`. That set the feature's ceiling: the label could only ever be a fragment wedged inside the `<code>` element, because that was the only place a string replacement could reach. Anything else a code block might want — a copy button, a header bar, a collapsed state — had nowhere to live. The class even carried a `@todo` saying as much.

Fenced code blocks now render through the publishable `components/markdown/code-block.blade.php`, which receives the rendered code block markup as `$contents`, along with `$language` and `$label`, and decides what goes around it. Block chrome becomes a view change rather than a framework change.

## How it works

Hyde renders the view around a code block, not the code inside it. Once the document is parsed and every listener has seen it, each remaining fence is wrapped in a `CodeBlock` node of Hyde's own, and it is that wrapper the renderer is registered for. The fence stays in the tree as its only child and is rendered through the child renderer, which dispatches it back through the same environment — where Hyde is not a candidate, because Hyde renders `CodeBlock`, not `FencedCode`. So whichever renderer the environment already had for a fence is the one that still renders it, exactly once, whether that is CommonMark's own renderer, Torchlight, or a third-party extension. Renderer priority is not part of the contract, and no highlighter is named in the framework wiring or depended on.

## Breaking: `title="…"` replaces the filepath comment

Code block labels are set with a `title="…"` modifier on the fence, using the same attribute syntax as terminal block titles:

````markdown
```php title="app/Models/Example.php"
echo 'Hello World!';
```
````

The v2 `// filepath:` comment syntax is removed, along with its `#`, `/* */`, and `<!-- -->` variants. A comment left behind is no longer recognized, so it stays in the code as an ordinary first line rather than being silently dropped.

Two syntaxes were not worth carrying. The comment's one real advantage was portability — it still conveys the file when the same Markdown is read somewhere that knows nothing about Hyde — but it also says the label must be a file path when it is really a title, puts presentation metadata inside the displayed code, needs a comment variant per language, makes Hyde delete a line that could be legitimate first-line code, and requires precedence rules once both forms exist. Degrading to a plain code block elsewhere is acceptable for framework-specific presentation metadata, which is how `title="…"` already behaves for terminal blocks.

The label is no longer tied to file paths, so a block can be titled with anything. The language is optional, so ` ``` title=".env" ` labels a block that declares none, which is then treated as `plaintext`.

Converting existing blocks is mechanical, and belongs to the v3 upgrade script rather than the framework. The rules it should follow are recorded in `HYDEPHP_V3_PLANNING.md`, and the upgrade guide describes the manual replacement until that script exists. This repository's own documentation was migrated the same way, and all 87 labels in it converted cleanly.

## Breaking: the filepath label view was removed

`components/filepath-label.blade.php` was removed. The label markup now lives in `components/markdown/code-block.blade.php`, so published label customizations must be moved into that view.

The upgrade guide covers the steps: publish the code block view, reapply customizations, delete the old view, and check custom CSS. It points at `php hyde publish:views` rather than `publish:views components`, since passing the group name overwrites every published component view.

The generated markup around code blocks has changed. `hyde-code-block` and `hyde-code-block-label` are stable hooks for your own CSS. Syntax highlighting is unaffected.

## Blade blocks

Titled `blade` fences are treated as labelled code samples, while `blade render` and `blade component="name"` remain executable.

## Verification

- New `CodeBlocksTest` and `CodeBlockViewModelUnitTest`, plus a Code Blocks section in `ComposableMarkdownBlocksDocumentationTest` covering the documented code-block behaviour and view contract.
- The removed syntax is locked down: an old filepath comment renders no label and stays in the code verbatim, blank line included.
- Torchlight is covered against a faked API, and third-party fenced-code renderers against fixture extensions, including ones that keep state between listener and renderer, decline a block, register at high or low priorities, or replace the node outright.
- Full suite green, and HydeStan reports no errors.

Design rationale, including the approaches this went through before landing here, is recorded in `HYDEPHP_V3_PLANNING.md`.

### CODE REVIEW

File: docs/digging-deeper/composable-markdown-blocks.md [hunk 1]
```diff
@@ -670,7 +689,7 @@ Coloured blockquotes are expanded line by line, before the Markdown parser runs.
 invisible `U+200E LEFT-TO-RIGHT MARK` character — it's the standard workaround for documenting the syntax without
 triggering it.
```
emmadesilva, line 665: This has no reason being here. Will remove in future PR.

docs/digging-deeper/composable-markdown-blocks.md
```diff
@@ -689,11 +708,11 @@ A published view is a copy, frozen at the version you published it from. Framewo
...
 
-### Torchlight changes the markup
+### Your highlighter decides what `$contents` looks like
 
-When Torchlight is enabled, code blocks are rendered by Torchlight rather than by CommonMark's default renderer. The
-filepath label accounts for this via `$highlightedByTorchlight`, but a custom view that assumes one structure will look
-wrong under the other. Test both if your site toggles Torchlight.
+Torchlight and CommonMark's default renderer produce different markup for the same block, and a third-party
+highlighter produces something different again. Your view receives whichever one as `$contents`, so a custom view that
+digs into that markup expecting one structure will look wrong under another. Test both if your site toggles Torchlight.
```
emmadesilva, line 689: Probably also not relevant. Makes it seem like we offload highlighting to consumer, which we don't. We just support it.

PR #2587, opened 2026-08-11
[3.x] Support custom media directories in the realtime compiler

emmadesilva:
## Summary

- pass the resolved media source and output directories from `hyde serve` to the server process as `HYDE_SERVER_MEDIA_DIRECTORY` and `HYDE_SERVER_MEDIA_OUTPUT_DIRECTORY`, using the existing environment plumbing
- read those in `AssetFileLocator` instead of hardcoding `_media` and `media/`, so a project using the `media_directory` config option gets its media served instead of 404s
- share the resolved output directory with the router's pre-boot fast path, keeping media requests off the framework bootstrap
- resolve dashboard media preview links through `Hyde::getMediaOutputDirectory()`, which was hardcoded to `media/` as well

Requests that require the framework boot a fresh application, so resolved configuration cannot be shared with the pre-boot media fast path through application memory. The server process environment is where it can live instead, without paying for a bootstrap on every asset request. The defaults still apply when the server is started directly, such as through the Herd integration, where there is no serve command to resolve them.

Changing the media directory while the server is running is not picked up, since the values are resolved at startup. Restarting the server is the expected response, and this is not documented.

## Tests

- realtime compiler suite: 48 passed (146 assertions), including new end-to-end coverage for serving a custom media directory and for dashboard preview links; both were confirmed to fail against the unmodified sources
- `UnitFramework`: 1,922 passed, with the documented PHP 8.5 `FeaturedImageUnitTest` failure remaining
- `ServeCommand` unit and feature tests: 53 passed, including a regression test that a custom media directory is resolved into the server environment
- `php monorepo/HydeStan/run.php`: no errors


emmadesilva:
• QA result: PASS — no issues found.

  - Focused tests: 99 passed; the socket-dependent Vite test also passed when rerun with network permission.
  - Custom _qa-media directory served correctly at /qa-media/..., including nested files.
  - Custom assets directory worked without an underscore.
  - Missing assets and the old /media path returned 404.
  - Dashboard generated the correct custom media link.
  - hyde build copied assets to the configured output directory.
  - Direct server fallback continued serving default /media/app.css.
  - HydeStan reported no errors.
  - Worktree is clean; temporary QA files were removed.

─ Worked for 5m 33s ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  gpt-5.6-luna xhigh

